Issue #61, opened 2017-05-19
Await file open before attempting writes


ricellis:
When using the `--output` option the `couchbackup.bin.js` opens a writable stream for a provided filename, but does not wait for the `open` event before proceeding. Whilst in most cases the fetching of documents from the database should provide ample time for the `fs` to actually open the file before any writes are attempted it seems prudent to shut this potential window and avoid the potential `ENOENT` event on a write before the file is created.

FWIW I haven't actually seen this produce an error.

- couchbackup/couchrestore version(s) that are affected by this issue.
> 1.4.0, 2.0.0-SNAPSHOT
- Node.js version.
> Any
- A sample of the code or command line statement that demonstrates the issue.
```js
ws = fs.createWriteStream(config.COUCH_OUTPUT, { flags: flags });
// there is no ws.on('open', ...) or similar here
```

PR #95, opened 2017-06-05
Throw error when write fails

tomblench:
Related to #61, this PR addresses cases where we're writing to a stream and not a file, so it's not possible to check if it is writable before starting. Instead, make sure that the callback is called with the correct error

### DIFF
--- a/citest/test.js
+++ b/citest/test.js
@@ -322,3 +322,26 @@ describe('Event tests', function() {
     });
   });
 });
+
+describe('Write error tests', function() {
+  it('calls callback with error set when stream is not writeable', function(done) {
+    u.timeoutFilter(this, 10);
+    const dirname = fs.mkdtempSync('test_backup');
+    // make temp dir not writeable
+    fs.chmodSync(dirname, 0);
+    const filename = dirname + '/test.backup';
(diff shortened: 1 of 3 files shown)


citest/test.js
```diff
@@ -322,3 +322,25 @@ describe('Event tests', function() {
...
+    fs.chmodSync(dirname, 0);
+    const filename = dirname + '/test.backup';
+    const backupStream = fs.createWriteStream(filename, {flags: 'w'});
+    const params = {useApi: true};
+    // try to do backup and check err was set in callback
+    u.testBackup(params, 'animaldb', backupStream, function(err) {
+      // error should have been set
+      assert.ok(err != null);
+      assert.equal(err.code, 'EACCES');
+      // cleanup temp dir
```
ricellis: The asserts will throw if they fail which will cause an `uncaught` exception in the tests.
This style worked more cleanly for me and made sure that the after hooks etc were not interrupted.
```js
try {
  assert ...
  done()
} catch (err) {
  done(err);
}
```
You will also probably want to move the cleanup before the asserts, otherwise it won't run in cases where the test fails.

ricellis:
Also needs a CHANGES entry.

PR #96, opened 2017-06-05
79 fix unreliable network tests

### LINKED ISSUE #79: Fix unreliable network tests
Implement necessary request retrying to ensure backup/restore completes successfully over an unreliable network.

smithsz:
## What
Add following poisons to toxy tests:
- bandwidth-limit
- latency
- slow-read

The remaining poisons (namely `abort`, `rate-limit` & `unexpected-errors`) will be implemented in a future PR as additional retrying logic is required.

_'[test] make test release build' commit to be removed prior to merge._

## How
Uncomment poisons and introduce TEST_TIMEOUT_MULTIPLIER to ensure toxy tests don't timeout.

## Issues
Fixes #79.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,59 +24,59 @@ def getEnv(envName) {
 
   // Add test suite specific environment variables
   switch(envName) {
-  case 'test-default':
-    envVars.add("COUCH_URL=https://${env.DB_USER}:${env.DB_PASSWORD}@${env.DB_USER}.cloudant.com")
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "couchbackup",
-  "version": "2.0.0-SNAPSHOT",
+  "version": "2.0.0",
```
ricellis: After the publishing branch merges, this is no longer a suitable way of testing as it will actually cause a release to happen.

PR #97, opened 2017-06-06
Run backup on file open

tomblench:
See #61

### DIFF
--- a/bin/couchbackup.bin.js
+++ b/bin/couchbackup.bin.js
@@ -47,7 +47,8 @@ if (program.output) {
   if (program.log && program.resume) {
     flags = 'a';
   }
-  ws = fs.createWriteStream(program.output, { flags: flags });
+  const fd = fs.openSync(program.output, flags);
+  ws = fs.createWriteStream(null, { fd: fd });
 }
 
 return couchbackup.backup(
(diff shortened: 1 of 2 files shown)


bin/couchbackup.bin.js
```diff
@@ -48,17 +48,25 @@ if (program.output) {
     flags = 'a';
   }
   ws = fs.createWriteStream(program.output, { flags: flags });
+  ws.on('open', function(fd) {
+    return backup();
```
ricellis: Does this actually work, or does it returned `undefined`?

We need* to return `backup` from the main of this script. This I think does that only for the `else` case. I think for the `output` case this is registering a callback that will return `backup`, but the main will register the event listener with that callback and then proceed to return `undefined`

* I think in practice we don't actually use value returned, so it does work by virtue of `backup` getting called in both cases, but I don't think it works strictly in the way it might appear to and I think it might be better to ensure that the same thing is returned in the same way for all cases.

I think it is possible to make the open synchronous instead:
```js
const fileDescriptor = fs.openSync(program.output, { flags: flags });
ws = fs.createWriteStream(null, {fd: fileDescriptor});
...
```

PR #99, opened 2017-06-06
Fix install and running of example scripts

### LINKED ISSUE #98: Scripts in `examples` don't work out of the box
There is a bug in the scripts in the examples folder. The install instructions are also not correct.

- Variable `url` shadows package `url`.
- Install instructions assume published couchbackup 2.0.0.

mikerhodes:
Fixes the minor bugs in the scripts in `examples`.

Fixes #98

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -4,7 +4,6 @@
   "description": "Examples of using CouchBackup as a library",
   "dependencies": {
     "aws-sdk": "^2.39.0",
-    "couchbackup": "^2.0.0",
     "tmp": "^0.0.31",
     "verror": "^1.10.0",
     "yargs": "^7.0.2"
--- a/examples/s3-backup-file.js
+++ b/examples/s3-backup-file.js
@@ -182,8 +182,8 @@ function uploadNewBackup(s3, backupTmpFilePath, bucket, key) {
  *
  * @param {string} url URL to safen
  */
-function s(url) {
-  var parts = url.parse(url);
+function s(originalUrl) {
+  var parts = url.parse(originalUrl);
   delete parts.auth;
   return url.format(parts);
 }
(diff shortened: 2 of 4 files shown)


examples/s3-backup-file.js
```diff
@@ -182,8 +182,8 @@ function uploadNewBackup(s3, backupTmpFilePath, bucket, key) {
  *
  * @param {string} url URL to safen
  */
-function s(url) {
-  var parts = url.parse(url);
+function s(originalUrll) {
```
ricellis: originalUrll -> originalUrl

examples/s3-backup-file.js
```diff
@@ -182,8 +182,8 @@ function uploadNewBackup(s3, backupTmpFilePath, bucket, key) {
...
+  var parts = url.parse(originalUrll);
```
ricellis: and again

examples/package.json
```diff
@@ -4,7 +4,6 @@
   "description": "Examples of using CouchBackup as a library",
   "dependencies": {
     "aws-sdk": "^2.39.0",
-    "couchbackup": "^2.0.0",
```
ricellis: Without this dependency here does the `require` statement need to be updated to point to the local file or it will it safely pick it up from the local `node_modeules` ahead of any globally installed copy?

mikerhodes: It picked it up for me from node_modules after `npm install ../`. I guess it's probably possible to just require the library directly, but didn't think of it before now.

examples/s3-backup-file.js
mikerhodes: Bah. This is an infuriating bug/behaviour of the VSCode vim plugin.

Issue #22, opened 2017-05-05
Abort processes for permanent errors


ricellis:
- couchbackup/couchrestore version(s) that are affected by this issue.
>1.4.0
- Node.js version.
>4.8.2
- A sample of the code or command line statement that demonstrates the issue.

Following on from #17 there are other cases where the tool will persist in streaming a large file with no hope of success. For example the case of attempting a restore without `_writer` permission for the database.

There are also cases where this applies to the backup process, for example, having `_reader` permissions revoked (if `_reader` was not there in the first place it would never spool the `_changes` and terminate before the batch get stage where the process currently tries forever to complete.

ricellis:
I think at the moment we still don't permanently fail if there is a `401` or `403` when trying to backup so I'll take another look at this.

PR #108, opened 2017-06-08
Improve error handling

tomblench:
See #22 

**What**

Abort processing for fatal errors when doing restore:

- Throw fatal error if the candidate restore database doesn't exist (HEAD returns 404)
- Throw fatal error if unexpected return code is obtained after doing HEAD on candidate restore database
- Throw fatal error if a 401 is encountered when trying to write to the database
- Stop processing queue and other threads if fatal errors are encountered

**How**

Make `writer` emit `error` events and then call the error callback as appropriate.

Make `exists` more robust.

**Testing**

Manual testing restoring to a database without `_writer` permissions:
```
$ COUCH_URL=https://tomblench.cloudant.com COUCH_DATABASE=airportdb_backup  ./bin/couchrestore.bin.js < out 
...
Database https://tomblench.cloudant.com/airportdb_backup does not have correct permissions. Check that you have the correct permssions before restoring.
```

Manual testing restoring to a database which does not exist:
```
COUCH_URL=https://tomblench.cloudant.com COUCH_DATABASE=nonexistent  ./bin/couchrestore.bin.js < out 
...
Database https://tomblench.cloudant.com/nonexistent does not exist. Create the target database before restoring.
```

Manual testing restoring to a database without `_reader` permissions:
```
COUCH_URL=https://tomblench.cloudant.com COUCH_DATABASE=airportdb_backup  ./bin/couchrestore.bin.js < out 
...
HEAD https://tomblench.cloudant.com/airportdb_backup returned error code 401
```

### DIFF
--- a/app.js
+++ b/app.js
@@ -201,20 +201,22 @@ module.exports = {
         if (err) {
           callback(err, null);
         }
-
-        writer.on('restored', function(obj) {
-          debug(' restored ', obj.total);
-          ee.emit('restored', {documents: obj.documents, total: obj.total});
-        })
-        .on('error', function(e) {
-          debug(' error', e);
-          ee.emit('error', e);
-        })
-        .on('finished', function(obj) {
-          debug('restore complete');
-          ee.emit('finished', {total: obj.total});
-          callback(null, obj);
-        });
+        if (writer != null) {
+          writer.on('restored', function(obj) {
+            debug(' restored ', obj.total);
+            ee.emit('restored', {documents: obj.documents, total: obj.total});
+          })
+          .on('error', function(e) {
+            debug(' error', e);
+            ee.emit('error', e);
+            callback(e, writer);
+          })
+          .on('finished', function(obj) {
+            debug('restore complete');
+            ee.emit('finished', {total: obj.total});
+            callback(null, obj);
+          });
+        }
       }
     );
 
--- a/includes/restore.js
+++ b/includes/restore.js
@@ -17,7 +17,9 @@ const request = require('./request.js');
 
 module.exports = function(dbUrl, buffersize, parallelism, readstream, callback) {
   exists(dbUrl, function(err, exists) {
-    if (err || !exists) {
+    if (err) {
+      callback(err, null);
+    } else if (!exists) {
       var e = new Error(`Database ${dbUrl} does not exist. ` +
         'Create the target database before restoring.');
       e.name = 'RestoreDatabaseNotFound';
@@ -51,9 +53,16 @@ function exists(dbUrl, callback) {
       callback(err, false);
       return;
     }
-    if (res && res.statusCode !== 200) {
-      callback(null, false);
+    if (res) {
+      if (res.statusCode === 200) {
+        callback(null, true);
+      } else if (res.statusCode === 404) {
+        callback(null, false);
+      } else {
+        // generic error code (for now, maybe map HTTP status codes to explanations?)
+        var e = new Error(`HEAD ${dbUrl} returned error code ${res.statusCode}`);
+        callback(e, false);
+      }
     }
-    callback(null, true);
   });
 }
--- a/includes/writer.js
+++ b/includes/writer.js
@@ -16,6 +16,11 @@
 const async = require('async');
 const request = require('./request.js');
 const stream = require('stream');
+const error = require('./error.js');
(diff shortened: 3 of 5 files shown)


includes/writer.js
```diff
@@ -47,6 +52,19 @@ module.exports = function(couchDbUrl, bufferSize, parallelism) {
       }
       cb();
     });
+
+    response.on('response', function(resp) {
+      if (resp.statusCode === 401) {
```
smithsz: catch a `403` here too

includes/writer.js
```diff
@@ -47,6 +52,19 @@ module.exports = function(couchDbUrl, bufferSize, parallelism) {
...
+        var e = new error.BackupError('Unauthorized', `Database ${couchDbUrl} does not have correct permissions. Check that you have the correct permissions before restoring.`);
+        response.abort();
+        writer.end();
+        // only emit the first error as there are multiple threads
+        if (!didError) {
```
smithsz: I don't think you need this. Let all the writers emit `error`. We appear only to handle the first and terminate.

includes/restore.js
```diff
@@ -27,6 +29,11 @@ module.exports = function(dbUrl, buffersize, parallelism, readstream, callback)
     var liner = require('../includes/liner.js');
     var writer = require('../includes/writer.js')(dbUrl, buffersize, parallelism);
 
+    writer.on('error', function(err) {
+      writer.end();
```
smithsz: Is this needed?

includes/writer.js
```diff
@@ -47,6 +52,19 @@ module.exports = function(couchDbUrl, bufferSize, parallelism) {
       }
       cb();
     });
+
+    response.on('response', function(resp) {
+      if (resp.statusCode === 401) {
+        var e = new error.BackupError('Unauthorized', `Database ${couchDbUrl} does not have correct permissions. Check that you have the correct permissions before restoring.`);
+        response.abort();
+        writer.end();
```
smithsz: Is this needed?

includes/restore.js
```diff
@@ -27,6 +29,11 @@ module.exports = function(dbUrl, buffersize, parallelism, readstream, callback)
     var liner = require('../includes/liner.js');
     var writer = require('../includes/writer.js')(dbUrl, buffersize, parallelism);
 
+    writer.on('error', function(err) {
```
smithsz: Just eyeballing this again on the train, we already have a writer 'error' listener in app.js. I reckon we can remove this and make the existing one call `callback(e, null);`. What say?

includes/writer.js
tomblench: Without this check we get multiple error messages so it seems to be necessary

includes/writer.js
tomblench: Done in 85a2d6d

includes/restore.js
tomblench: Good idea, done in 5f0ddf9

includes/writer.js
smithsz: it's only just dawned on me that we silently fail all error codes here.
did you want to deal with this in the pr? or spin out to anther issue?

app.js
```diff
@@ -201,20 +201,22 @@ module.exports = {
...
-        });
+        if (writer != null) {
+          writer.on('restored', function(obj) {
+            debug(' restored ', obj.total);
+            ee.emit('restored', {documents: obj.documents, total: obj.total});
+          })
+          .on('error', function(e) {
+            debug(' error', e);
+            ee.emit('error', e);
+            callback(e, writer);
```
smithsz: As discussed, I can tidy this up in exception handling work.

includes/writer.js
tomblench: Ideally we should do the following:
- Distinguish between permanent (4xx) and transient (5xx) codes. The latter can be re-tried as and when we have re-try logic in place.
- Have a shared lookup table which translates HTTP status codes into human readable errors which make sense in the context of the tool. I think there is a lot of value in a message like `Database ${couchDbUrl} does not have correct permissions. Check that you have the correct permissions before restoring` rather than `Error 401`.

tomblench: I added 9f0e859 which catches the 403 and as a TODO to the effect of the above

PR #112, opened 2017-06-09
Limit backup /_bulk_get attempts

### LINKED ISSUE #111: Don't retry backup `/_bulk_gets` indefinitely
Currently the backup tool retries batches indefinitely if it doesn’t manage to fetch them.

Implement a max. retry limit and error out on failure.

smithsz:
## What
Add a max. retry limit of 3 for all backup '/_bulk_get' requests.

Since failed backup '/_bulk_get' requests now don't need re-adding to the work queue, we don't need to re-read the log file on every iteration.

## How
Use `async.retry` to retry '/_bulk_get' requests  on failure.

## Issues
Fixes #111 .

### DIFF
--- a/bin/couchbackup.bin.js
+++ b/bin/couchbackup.bin.js
@@ -57,7 +57,7 @@ return couchbackup.backup(
   opts,
   error.terminationCallback
 ).on('written', function(obj) {
-  debug('written', obj.batch, ' docs: ', obj.total, 'Time', obj.time);
+  debug('written batch:', obj.batch, 'total docs now written:', obj.total, 'time:', obj.time);
 }).on('error', function(e) {
   debug('ERROR', e);
 }).on('finished', function(obj) {
--- a/includes/backup.js
+++ b/includes/backup.js
@@ -22,6 +22,8 @@ const spoolchanges = require('./spoolchanges.js');
 const logfilesummary = require('./logfilesummary.js');
 const logfilegetbatches = require('./logfilegetbatches.js');
 
+var client;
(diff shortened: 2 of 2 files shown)


ricellis:
Spooling the changes for a 50 GB database with a reasonable amount of branching generates a log file of ~300 MB. I think the default memory for Node is 512 MB; consequently I think queuing all the batches for, say, a 100 GB database will almost certainly cause an OOM.

Whilst I agree with not re-reading the log file over and over, I do think that because of this we need to limit the number of batches that are queued from the log file at a time. Maybe we can use the `unsaturated` and `saturated` functions of the queue to control pausing and unpausing the reading of lines from the log file? A re-read would be required on a resume, but there is no way around that.

smithsz:
I'm not sure I follow. I was very careful to ensure we only queued 50 batches at a time (mimicking the current behaviour). We do fetch all batch IDs and store them in a list. But, as you said, that's unavoidable as we need to see all the log lines on a resume. I'll have another look over this. Perhaps I've missed something.

ricellis:
Possibly I misunderstood the reading of all the batch IDs, will also look again.
I'm cautious about this significant behaviour change close to the 2.0 milestone given it is time consuming to test many large backups to ensure there are no memory issues introduced. I think we need to add the larger backups to the automated testing, but I'm not quite sure how to toggle them effectively via Jenkins, since they are too slow for every build (or even master build).

smithsz:
Lets revisit this post-2.0.

PR #114, opened 2017-06-10
Fetch document revisions individually if bulk get isn't supported

### LINKED ISSUE #107: Detect bulk_get before attempting backup
Some servers are running older versions of Cloudant e.g. https://00d20a8e-8bd6-454e-9a9c-7920172653d3-bluemix.cloudant.com/

Some are running the latest: e.g. https://reader.cloudant.com/

The former does not work with `couchbackup` because it doesn't support ` POST /db/_bulk_get`, so the changes feed spooling works, but the actual backup does not.

Perhaps this tool should determine whether `POST /db/_bulk_get` exists before beginning the backup.

PouchDB does this by calling `GET /db/_bulk_get` and seeing if it gets a `HTTP/1.1 405 Method Not Allowed` response. If it gets 405, it assumes that the POST equivalent will work.

smithsz:
# What
Check for the `/_bulk_get` endpoint. If it's missing then fetch document revisions individually.

## How
Check the bulk get endpoint exists using a GET `/_bulk_get` request.
 - On `405 Method Not Allowed`, proceed as normal.
 - On `404 Object Not Found`, fetch documents individually using the GET `/db/docid` endpoint with a selection of URL parameters to ensure all live and non-live revisions are captured in the backup.

## Testing
Includes additional CI test for unsupported `/_bulk_get`.

## Issues
Fixes #107.

### DIFF
--- a/bin/couchbackup.bin.js
+++ b/bin/couchbackup.bin.js
@@ -57,7 +57,7 @@ return couchbackup.backup(
   opts,
   error.terminationCallback
 ).on('written', function(obj) {
-  debug('written', obj.batch, ' docs: ', obj.total, 'Time', obj.time);
+  debug('written batch:', obj.batch, 'total docs now written:', obj.total, 'time:', obj.time);
 }).on('error', function(e) {
   debug('ERROR', e);
 }).on('finished', function(obj) {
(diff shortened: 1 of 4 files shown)


smithsz:
If there aren't any live leaves then this only fetches the chosen non-live leaf, not all of them.

I'll rework the queue task logic to use `POST _revs_diff` instead.

smithsz:
Will rework this on top of master.

Issue #32, opened 2017-05-08
Clean up error handling


ricellis:
- couchbackup/couchrestore version(s) that are affected by this issue.
>1.4.0
- Node.js version.
>N/A
- A sample of the code or command line statement that demonstrates the issue.
>N/A

Refactor the error handling so that the API and internal functions only emit error events and do not write to `stderr`. This will avoid suppressing errors from being handled by API callers and also give API users complete control over the output.
The command line tooling (effectively an API user) should write to `stderr` when receiving the events. <-- I think this part was handled by #52 (but whoever picks this up should check)

We should also check for whether something calls the `callback(err)` or emits an `error` event and document the expected behaviour.

smithsz:
Closely related to #103. I'll pick this one up too.

smithsz:
After some discussion,
- an event emitter `error` will represent a non-fatal error
(will be printed to stderr in the CLI)
- a callback error will represent a fatal error
(will terminate the CLI with a suitable exit code)

PR #118, opened 2017-06-13
115 validate writer requests

### LINKED ISSUE #115: Validate bulk docs response codes in writer.js
Currently we silently ignore any write failures without retry.

https://github.com/cloudant/couchbackup/blob/master/includes/writer.js#L41

It would be nice to validate the bulk docs response too, ensuring it contains the expected ids/revs (and no error/reason).

smithsz:
## What
Use newly added `request.isOK` & `request.isFatal` HTTP status code validators anywhere we validate HTTP response codes.

## How
Add HTTP status code validators to `request.js`. 

- `request.isFatal` will return a new `HTTPFatalError` error on failure that will terminate the CLI.
- `request.isOK` will return a new `HTTPError` error on failure that logs an error with the event emitter.

## Issues
Fixes #115 and is closely related to #32.

### DIFF
--- a/includes/request.js
+++ b/includes/request.js
@@ -17,9 +17,21 @@ const pkg = require('../package.json');
 const http = require('http');
 const https = require('https');
 const request = require('request');
+const error = require('./error.js');
 
 const userAgent = 'couchbackup-cloudant/' + pkg.version + ' (Node.js ' +
       process.version + ')';
+
+var checkResponse = function(resp, errName) {
+  errName = errName || 'HTTPError';
+  if (resp.statusCode >= 400) {
+    var errMsg = `${resp.statusCode} ${resp.statusMessage || ''}: ${resp.request.method} ${resp.request.uri.href}`;
+    if (resp.body && resp.body.error && resp.body.reason) {
+      errMsg += ` - Error: ${resp.body.error}, Reason: ${resp.body.reason}`;
+    }
+    return new error.BackupError(errName, errMsg);
+  }
+};
 
 module.exports = {
   client: function(url, parallelism) {
@@ -34,5 +46,11 @@ module.exports = {
       headers: {'User-Agent': userAgent},
       json: true,
       gzip: true});
+  },
+  checkResponseAndCallbackError: function(resp, callback) {
+    callback(checkResponse(resp));
+  },
+  checkResponseAndCallbackFatalError: function(resp, callback) {
+    callback(checkResponse(resp, 'HTTPFatalError'));
   }
 };
--- a/includes/spoolchanges.js
+++ b/includes/spoolchanges.js
@@ -18,6 +18,7 @@ const fs = require('fs');
 const liner = require('./liner.js');
 const change = require('./change.js');
 const error = require('./error.js');
+const debug = require('debug')('couchbackup:spoolchanges');
 
 /**
  * Write log file for all changes from a database, ready for downloading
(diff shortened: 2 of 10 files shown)


includes/request.js
```diff
@@ -34,5 +46,11 @@ module.exports = {
       headers: {'User-Agent': userAgent},
       json: true,
       gzip: true});
+  },
+  isOk: function(resp, callback) {
```
ricellis: ok so these names are a bit confusing, I was under the impression they were booleans.
My suggestions are terribly wordy but I think they articulate the purpose, perhaps there is a suitable compromise :)
`checkResponseAndCallbackError`
`checkResponseAndCallbackFatalError`

includes/spoolchanges.js
```diff
@@ -71,24 +71,28 @@ module.exports = function(dbUrl, log, bufferSize, callback) {
...
-        callback(new error.BackupError('SpoolChangesError', `ERROR: Changes request terminated before last_seq was sent`));
+    request.isFatal(resp, function(err) {
+      if (err) {
+        c.abort();
+        callback(err);
       } else {
-        logStream.write(':changes_complete ' + lastSeq + '\n');
-        logStream.end();
-        callback();
+        console.error('Streaming changes to disk:');
```
ricellis: Should be a `debug` line, although I realise it didn't really change in this PR.

includes/spoolchanges.js
```diff
@@ -71,24 +71,28 @@ module.exports = function(dbUrl, log, bufferSize, callback) {
...
+        resp
+          .pipe(liner())
+          .pipe(change(onChange))
+          .on('finish', function() {
+            processBuffer(true);
+            console.error('');
```
ricellis: I'm not sure how much impact this has on the prettiness of our output, but I think it should be removed. If we need a line break we should add a `\n` somewhere appropriate.

includes/writer.js
```diff
@@ -54,26 +53,16 @@ module.exports = function(couchDbUrl, bufferSize, parallelism) {
...
-          e = new error.BackupError('Forbidden', `Incorrect credentials for database ${couchDbUrl}. Check that you have the correct username and password or API key before restoring.`);
-          break;
-      }
-      if (e != null) {
-        response.abort();
-        // only emit the first error as there are multiple threads
-        if (!didError) {
-          didError = true;
-          writer.emit('error', e);
+      request.isOk(resp, function(err) {
```
ricellis: I've lost the plot a little since the changes while I was out, but does this actually retry like backup does? I had a feeling it didn't and if it doesn't shouldn't this be fatal, on the basis that some docs will now be missing from the restored DB?

includes/request.js
smithsz: Makes sense. I'll change.

includes/spoolchanges.js
smithsz: Erm, no idea, I just left it in from before. I'll see how bad it looks without or add the `\n`.

includes/writer.js
smithsz: You're right. There's no retrying here. This should be a fatal error and terminate execution. At the moment it just spits an error out to stderr. I'll change this.

smithsz:
We've only got the E2E testing as stands. I'd like to look at introducing something like [nock](https://github.com/node-nock/nock) so we don't have to rely of toxy for testing these edge cases.

includes/spoolchanges.js
```diff
@@ -18,6 +18,7 @@ const fs = require('fs');
 const liner = require('./liner.js');
 const change = require('./change.js');
 const error = require('./error.js');
+const debug = require('debug')('couchbackup:writer');
```
ricellis: Needs to be `couchbackup:spoolchanges`

ricellis:
@smithsz Nock tests look fine for asserting the response failure callbacks.

PR #119, opened 2017-06-13
Retry spool changes on failure

### LINKED ISSUE #100: Retry backup spool changes request on failure
A failure in the spool changes request will terminated the backup process.
https://github.com/cloudant/couchbackup/blob/master/includes/spoolchanges.js#L74

We should retry this request on failure using a suitable `since` parameter.

smithsz:
## What
Retry changes request on failure (using `since` parameter).

## How
Use `async.retry` to retry the changes spoolers on failure. Failures include,
- missing `last_seq`
- network error
- bad HTTP status code

## Testing
Includes additional CI tests.

## Issues
Fixes #100.


ricellis:
I did a bit of experimentation with the `seq_interval` and while it does offer a performance boost it does not seem significant across the range of batch sizes we might expect (e.g. it makes no significant difference to how long it takes to spool changes using `seq_interval=500` or `seq_interval=10000`), but 500 is significantly faster than having it enabled for every change.

I think it would be good to change the `seq_interval` to match the buffer size and then the retries can be for a single batch and there should be a `seq` for every completed batch, allowing the resume code to be simpler. WDYT?

smithsz:
This is still a little tricky as the first non-null `seq` value can often be offset by an arbitrary number of changes. I think the first batch should be of size <= bufferSize in order to absorb this offset and line us up with the `seq_interval`.

Lets revisit this post-2.0.

PR #123, opened 2017-06-16
Verify /_bulk_get endpoint is supported

smithsz:
## What
Verify database supports `/_bulk_get` endpoint prior to running backup.

## How
Query `GET /_bulk_get`:
- 404 => endpoint is _not_ supported
- 405  => endpoint is supported

## Testing
No additional tests added.

## Issues
See #107.

### DIFF
--- a/includes/backup.js
+++ b/includes/backup.js
@@ -22,6 +22,8 @@ const spoolchanges = require('./spoolchanges.js');
 const logfilesummary = require('./logfilesummary.js');
 const logfilegetbatches = require('./logfilegetbatches.js');
 
+var client;
+
 /**
  * Read documents from a database to be backed up.
  *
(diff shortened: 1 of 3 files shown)


includes/backup.js
```diff
@@ -43,24 +45,51 @@ module.exports = function(dbUrl, blocksize, parallelism, log, resume) {
...
-      if (err) {
-        ee.emit('error', err);
-      } else {
-        downloadRemainingBatches(log, dbUrl, ee, start, batchesPerDownloadSession, parallelism);
-      }
-    });
-  }
+  client = request.client(dbUrl, parallelism);
+
+  validateBulkGetSupport(dbUrl, function(err) {
```
ricellis: I wonder if it is worth breaking out the actual backup piece into a named function so that it is a bit easier to follow what is happening with all the callbacks.
```js
validateBulkGetSupport(dbUrl, function(err) {
  if(err) {
    return ee.emit('error', err);
  } else {
    proceedWithBackup(); // <--- contains all the if(resume), spoolchanges, download etc
  }
```

PR #124, opened 2017-06-16
Remove stderr logging from library

smithsz:
## What
Remove stderr logging from library.

## How
Ensure all errors are passed via the event emitter to the listening client rather than being printed directly to stderr.

## Issues
See #32.

### DIFF
--- a/bin/couchbackup.bin.js
+++ b/bin/couchbackup.bin.js
@@ -51,15 +51,19 @@ if (program.output) {
   ws = fs.createWriteStream(null, { fd: fd });
 }
 
+debug('Fetching all database changes...');
+
 return couchbackup.backup(
   databaseUrl,
   ws,
   opts,
   error.terminationCallback
-).on('written', function(obj) {
-  debug('written', obj.batch, ' docs: ', obj.total, 'Time', obj.time);
+).on('changes', function(batch) {
+  debug('Total batches received:', batch + 1);
+}).on('written', function(obj) {
+  debug('Written batch ID:', obj.batch, 'Total document revisions written:', obj.total, 'Time:', obj.time);
 }).on('error', function(e) {
   debug('ERROR', e);
 }).on('finished', function(obj) {
-  debug('finished', obj);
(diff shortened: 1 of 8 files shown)


bin/couchbackup.bin.js
```diff
@@ -51,15 +51,24 @@ if (program.output) {
...
+
 return couchbackup.backup(
   databaseUrl,
   ws,
   opts,
   error.terminationCallback
-).on('written', function(obj) {
-  debug('written', obj.batch, ' docs: ', obj.total, 'Time', obj.time);
+).on('changes', function(batch) {
+  process.stderr.write(`\r  Total batches received: ${batch + 1}`);
```
ricellis: I'm not sure if it is worth over-writing this line, we don't overwrite when receiving the written events. I know you have just kept things in line with what was here previously, but I'm not sure it is necessary. Why not just replace this and the `firstWrite` stuff with a `debug` statement for each change line?

includes/writer.js
```diff
@@ -146,11 +148,11 @@ module.exports = function(couchDbUrl, bufferSize, parallelism) {
             done();
           });
         } else {
-          console.error('ERROR on line', linenumber, ': not an array');
+          ee.emit('error', new error.BackupError('CorruptBackup', `Error on line ${linenumber} of backup file - not an array`));
           done();
         }
       } catch (e) {
-        console.error('ERROR on line', linenumber, ': cannot parse as JSON');
+        ee.emit('error', new error.BackupError('CorruptBackup', `Error on line ${linenumber} of backup file - cannot parse as JSON`));
```
ricellis: I wish I could think of a better word than `Corrupt` to use here. Whilst there is a corrupted batch in the file if this error is thrown, as you know it is non-fatal and could have no impact on the completeness of what is being restored if it was a resumed backup.

ricellis: maybe just `BackupFileJsonError`

ricellis:
Thinking about this a bit more, I think the new `changes` events need to be documented somewhere, we formalized the event names in doc at some point.

PR #129, opened 2017-06-23
128 handle fatal errors in shallow backup

### LINKED ISSUE #128: Shallow backup does not handle fatal errors
The shallow backup mode needs to use the same error handling as the full backup. For example terminating for permanent HTTP problems (e.g. 401, 403).

Most of the functionality can come from the additional work I'm doing for #22, but the `shallowbackup.js` will need to call it.

- couchbackup/couchrestore version(s) that are affected by this issue.
>2.0.0_SNAPSHOT
- Node.js version.
>Any
- A sample of the code or command line statement that demonstrates the issue.

smithsz:
## What
Handle fatal errors in shallow backup.

## How
Use `request.checkResponseAndCallbackError` to check `/_all_docs` responses. If fatal, terminate the backup and emit an error.

## Testing
Includes additional unit tests.

The commented test requires https://github.com/cloudant/couchbackup/tree/22-permanent-errors.

## Issues
Fixes #128.

### DIFF
--- a/includes/shallowbackup.js
+++ b/includes/shallowbackup.js
@@ -14,61 +14,86 @@
 'use strict';
 
 const async = require('async');
+const error = require('./error.js');
+const events = require('events');
 const request = require('./request.js');
 
-module.exports = function(dbUrl, blocksize, parallelism, log, resume) {
-  const client = request.client(dbUrl, parallelism);
-  if (typeof blocksize === 'string') {
-    blocksize = parseInt(blocksize);
-  }
-  const events = require('events');
-  const ee = new events.EventEmitter();
+// Both the 'log' and 'resume' parameters are unused in this function. They
+// exist so that the method signatures for shallow backup and backup are
+// symmetrical.
+module.exports = function(dbUrl, limit, parallelism, log, resume) {
+  if (typeof limit === 'string') limit = parseInt(limit);
 
-  var startdocid = null;
+  const client = request.client(dbUrl, parallelism);
+  const ee = new events.EventEmitter();
   const start = new Date().getTime();
-  var batch = 1;
+  var batch = 0;
+  var hasErrored = false;
+  var startKey = null;
   var total = 0;
 
-  async.doUntil(function(callback) {
-    var opts = {limit: blocksize + 1, include_docs: true};
-    if (startdocid) {
-      opts.startkey_docid = startdocid;
-    }
-    var r = {
-      url: dbUrl + '/_all_docs',
-      method: 'GET',
-      qs: opts
-    };
-    client(r, function(err, res, data) {
-      if (err || !data.rows) {
-        ee.emit('error', err || data);
-        return callback(null, null);
-      }
+  async.doUntil(
+    function(callback) {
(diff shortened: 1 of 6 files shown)


test/shallowbackup.js
```diff
@@ -0,0 +1,140 @@
...
+
+/* global describe it */
+'use strict';
+
+const assert = require('assert');
+const backup = require('../includes/shallowbackup.js');
+const fs = require('fs');
+const nock = require('nock');
+
+describe('#unit Perform backup using shallow back', function() {
```
ricellis: shallow back....
up!

test/shallowbackup.js
```diff
@@ -0,0 +1,140 @@
...
+  const dbUrl = 'http://localhost:5984/animaldb';
+
```
ricellis: I like to see a `beforeEach` that does `nock.cleanAll()` I found it very helpful when faced with multiple mock failures in a single suite.

includes/shallowbackup.js
```diff
@@ -14,61 +14,78 @@
 'use strict';
 
 const async = require('async');
+const error = require('./error.js');
+const events = require('events');
 const request = require('./request.js');
 
-module.exports = function(dbUrl, blocksize, parallelism, log, resume) {
+module.exports = function(dbUrl, limit, parallelism, log, resume) {
```
ricellis: I think we should have a comment here regarding the signature parameters that are unused.

test/shallowbackup.js
```diff
@@ -0,0 +1,140 @@
...
+      // batch 3
+      .get('/_all_docs')
+      .query({limit: 3, include_docs: true, startkey: '"kookaburra\\u0000"'})
+      .reply(200, JSON.parse(fs.readFileSync('./test/fixtures/animaldb_all_docs_3.json', 'utf8')))
+      // batch 4
+      .get('/_all_docs')
+      .query({limit: 3, include_docs: true, startkey: '"snipe\\u0000"'})
+      .reply(200, JSON.parse(fs.readFileSync('./test/fixtures/animaldb_all_docs_4.json', 'utf8')));
+
+    backup(dbUrl, 3, 1, null, false)
```
ricellis: maybe call this with `dbUrl, 3, null, null, null` since the args are ignored anyway, this makes it look like they have purpose
and for the later instances of the same call

includes/shallowbackup.js
```diff
@@ -14,61 +14,78 @@
...
-    };
-    client(r, function(err, res, data) {
-      if (err || !data.rows) {
-        ee.emit('error', err || data);
-        return callback(null, null);
-      }
+  async.doUntil(
+    function(callback) {
+      var opts = {limit: limit, include_docs: true};
+      if (startKey) opts.startkey = `"${startKey}\\u0000"`; // exclusive start key
```
ricellis: This could probably use a comment, as it is a bit of trick.
> To avoid double fetching a document solely for the purposes of getting the next ID to use as a `startkey` for the next page we instead use the last ID of the current page and append the lowest unicode sort character.

PR #150, opened 2017-07-21
Migrated to use nodejs-cloudant

ricellis:

## What

Changed to use nodejs-cloudant for database requests.

## How

* Moved `nodejs-cloudant` package from `devDependencies` to `dependencies` and updated to latest version.
* Refactored `request.js` to return a nodejs-cloudant `db` scope for making requests.
* Updated uses of `request.js` to use the nodejs-cloudant `db` scope methods.
* Removed URL from internal classes and passed nodejs-cloudant `db` instance instead.
* Nodejs-cloudant performs some request checking and calls back with an error object, so removed `request.js` checkResponse* methods.
* Added `error` `convertResponseError` and `convertResponseErrorToFatal` methods for turning nodejs-cloudant errors into backup errors.
* Added error message augmentation to glean more detail than Nano provides.
* Deleted unnecessary `command` parameter from `_bulk_get` payload.
* Refactored `_bulk_get` check to use `HEAD` instead of `GET`.
* Added extra detail for socket messages.

## Testing

* Refactored tests as necessary to use nodejs-cloudant instead of request.
* Added 2 new error test cases.
* Existing tests continue to pass using new dependency.

## Issues

None.


includes/error.js
```diff
@@ -63,10 +64,63 @@ class HTTPFatalError extends HTTPError {
...
+      return augmentMessage(err);
+    }
+  });
+}
+
+function augmentMessage(err) {
+  // For errors that don't have a status code, we are likely looking at a cxn
+  // error.
+  // Try to augment the message with more detail
+  // TODO add this extra message detail to nano?
```
smithsz: Can you raise an issue in cloudant-nano?

includes/backup.js
```diff
@@ -215,49 +222,43 @@ function processBatchSet(dbUrl, parallelism, log, batches, ee, start, grandtotal
...
-            total += output.length;
-            var t = (new Date().getTime() - start) / 1000;
-            ee.emit('received', {
-              batch: thisBatch,
-              data: output,
-              length: output.length,
-              time: t,
-              total: total
-            }, q, logCompletedBatch);
+    // TODO should use built-in _bulk_get, but revs not accepted as part of body
```
ricellis: I don't think we should add complexity to the client to workaround a service problem, better to make the Couch API consistent. If that doesn't happen then we could consider trying to update the nano method to accept both a qs and a body.

includes/error.js
ricellis: Done https://github.com/apache/couchdb-nano/issues/58

includes/request.js
```diff
@@ -59,12 +29,14 @@ module.exports = {
...
-  checkResponseAndCallbackError: checkResponseAndCallbackError,
-  checkResponseAndCallbackFatalError: checkResponseAndCallbackFatalError
+    // Split the URL for use with nodejs-cloudant
+    var actUrl = url.substr(0, url.lastIndexOf('/'));
+    var dbName = url.substr(url.lastIndexOf('/') + 1);
+    return cloudant({url: actUrl,
+      requestDefaults: {
+        agent: keepAliveAgent,
+        headers: {'User-Agent': userAgent},
+        // TODO this breaks nodejs-cloudant! json: true,
```
ricellis: Not yet. I'll try to understand a bit more about why it goes wrong.

ricellis: So I think it is basically too much `JSON.stringify` is going on. If `json: true` is set then request will stringify JSON objects set as the body, which Nano also does. There is a `dontStringify` option in Nano, but it appears to be an internal thing.
I don't think this is necessarily a bug other than perhaps the fact that it is undocumented that you shouldn't combine these options, although I suppose Nano could check the request options and avoid the double stringification. I'm not sure what the impact on responses might be in that case.

I've removed this TODO in 7490074bd21f4526ae127b4876764d6f54a1257c since I don't think we need it if using nodejs-cloudant.

While debugging I noticed some duplicate `include_docs: true` options too, which I removed in f94afe689110d94567f8b0e955121b1c7b7f047f.

includes/backup.js
```diff
@@ -80,20 +78,28 @@ module.exports = function(dbUrl, blocksize, parallelism, log, resume) {
...
           return new error.BackupError('BulkGetError', 'Database does not support /_bulk_get endpoint');
         case 405:
           // => supports /_bulk_get endpoint
           return;
         default:
-          return new error.HTTPFatalError(res);
+          return new error.HTTPFatalError(err);
       }
     });
+    if (err) {
```
ricellis: Nope, removed in ed040130d562c4e002450af0ce6d4a2ae62e8de5

includes/backup.js
```diff
@@ -215,49 +218,43 @@ function processBatchSet(dbUrl, parallelism, log, batches, ee, start, grandtotal
...
-            total += output.length;
-            var t = (new Date().getTime() - start) / 1000;
-            ee.emit('received', {
-              batch: thisBatch,
-              data: output,
-              length: output.length,
-              time: t,
-              total: total
-            }, q, logCompletedBatch);
+    // TODO should use built-in _bulk_get, but revs not accepted as part of body
```
emlaver: Remove TODO - should this be replaced with `Note:` or similar?

includes/backup.js
```diff
@@ -80,20 +78,28 @@ module.exports = function(dbUrl, blocksize, parallelism, log, resume) {
  * @param {string} dbUrl - URL of database
```
ricellis: Fixed in 6efd6e7d722d650e2ac29e6fcbc592ccc9d71ff3

includes/backup.js
```diff
@@ -102,7 +108,7 @@ function validateBulkGetSupport(dbUrl, callback) {
  * to avoid enqueueing too many in one go.
  *
  * @param {string} log - log file name to maintain download state
- * @param {string} dbUrl - Source database URL
+ * @param {string} db - nodejs-cloudant db
  * @param {events.EventEmitter} ee - event emitter to emit received events on
  * @param {time} start - start time for backup process
```
ricellis: Unrelated to my changes, but yes it should. Fixed in 8cfff502b26d965df9c338652ba4648b460dc85c

includes/backup.js
ricellis: Fixed in 8aadbf9107f606778924824ac423fbb27013dd2e

Issue #159, opened 2017-09-29
Support using an IAM API key


ricellis:
Rebase on nodejs-cloudant to pick up other auth support.
Add configuration options.

PR #184, opened 2018-01-31
Changed retry logic to nodejs-cloudant

ricellis:
Thanks for your hard work, please ensure all items are complete before opening a PR.

- [x] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](https://github.com/cloudant/couchbackup/blob/master/DCO1.1.txt)
- [x] You have [added tests](https://github.com/cloudant/couchbackup/blob/master/CONTRIBUTING.md#testing) for any code changes
- [x] You have updated the [CHANGES.md](https://github.com/cloudant/couchbackup/blob/master/CHANGES.md)
- [x] You have updated the [.npmignore](https://github.com/cloudant/couchbackup/blob/master/.npmignore) _(if applicable)_ - *N/A*
- [x] You have completed the PR template below:

## What

Changed retry logic to nodejs-cloudant.

## How

Instead of handling retries internally, delegate to nodejs-cloudant's `retry` plugin.
i) Configured `retry` plugin.
ii) Removed transient `HTTPError` errors and associated events.
iii) Updated `README` for errors no longer being emitted.

*For consideration by reviewers*
The couchbackup API no longer emits `error` events for transient errors. It could be argued that this is a breaking change, but the previous code and this new version both retry the transient error anyway. It's possible (but doubtful given the lack of a functioning cancellation API) that users were cancelling backups on transient errors, which would not longer be possible.

## Testing

iv) Updated tests for appropriate default retry numbers.
v) Removed obsolete transient callback tests and merged with associated failure callback tests.

## Issues

Should prevent errors like #154 
Part of #173

PR #185, opened 2018-01-31
Iam support

ricellis:

## What

Added support for using an IAM API key.

## How

* Refactored to pass options dict instead of individual options:
    * Access to the `iamApiKey` and `iamTokenEndpoint` override are required in the `request` class places, so rather than extended the internal function signatures to keep adding more parameters it makes sense to just pass the entire options dictionary around now.
* Added iamApiKey option to CLI and opts:
    * Added `--iam-api-key` (`iamApiKey`) option.
    * Added documentation for new option.
* Added validation for URL auth and iam-api-key:
    * Prevent users supplying URL credentials and an IAM API key.
* Configured the nodejs-cloudant `iamauth` plugin when key supplied, default to the `cookieauth` plugin if not.
* Updated CHANGES.md.

## Testing

* Added basic tests for `iam-api-key` config.
* Added URL validation tests.
* Added IAM branch to parallel tests that runs the non-`#unit` non-`#slow` tests with an IAM API key.

## Issues

#159

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@
 - [IMPROVED] Enhanced resilience of backup and restore processes by enabling the
   nodejs-cloudant retry plugin.
 - [UPGRADED] Increased nodejs-cloudant dependency to 2.x.
+- [NEW] Added API for upcoming IBM Cloud Identity and Access Management support
+  for Cloudant on IBM Cloud. Note: IAM API key support is not yet enabled in the
+  service.
+- [IMPROVED] Added URL validation for presence of host and database elements.
 
 # 2.0.1 (2018-01-11)
 
--- a/app.js
+++ b/app.js
@@ -1,4 +1,4 @@
-// Copyright © 2017 IBM Corp. All rights reserved.
+// Copyright © 2017, 2018 IBM Corp. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -27,6 +27,7 @@ const events = require('events');
 const debug = require('debug')('couchbackup:app');
 const error = require('./includes/error.js');
 const fs = require('fs');
+const legacyUrl = require('url');
 
 /**
  * Test for a positive, safe integer.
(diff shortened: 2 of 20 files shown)


CHANGES.md
```diff
@@ -4,7 +4,11 @@
 # 2.0.1 (2018-01-11)
 
 - [NEW] Changed to use nodejs-cloudant for database requests.
+- [NEW] Added API for upcoming IBM Cloud Identity and Access Management support
+  for Cloudant on IBM Cloud. Note: IAM API key support is not yet enabled in the
+  service.
 - [IMPROVED] Added compression to restore process requests.
+- [IMPROVED] Added some URL validation.
```
emlaver: Instead of using 'some', maybe it could say something like "Added Cloudant URL and IAM API key validation".

app.js
```diff
@@ -82,6 +87,32 @@ function validateArgs(url, opts, cb) {
     cb(new error.BackupError('InvalidOption', 'Invalid resume option, must be type boolean'), null);
     return;
   }
+
+  // Validate URL and ensure no auth if using key
```
emlaver: Is it worth checking the length of host/path? i.e. `urlObject.host.length !== 0`

ricellis:
BTW I'll update the copyrights on this after it is re-based on top of #184.

app.js
ricellis: As well as `undefined` and `null` the empty string is `falsy` in javascript, so I don't think the length check adds anything.

CHANGES.md
ricellis: I updated this in 7680793533fcfa2c4d45d7256175a43eddde8537
`- [IMPROVED] Added URL validation for presence of host and database elements.`
I think this makes sense because that is the improvement from the previous version. The validation of the IAM API key itself is part of adding the support for IAM. Let me know if you agree.

app.js
ricellis: When I was looking at this though I noticed that the exception message was a little confusing so I updated it in 42af7450c99f47a337024f7359b99eb119fd68db

emlaver: Ok sounds good.

PR #223, opened 2018-05-18
Added DB existence check for backup

### LINKED ISSUE #212: Couchbackup should give a better error message when incorrect creds are used
Currently a zero length backup file results and the output is
```
couchbackup Backup complete - written{"total":0,"err":null} +0ms
```

ricellis:

## What

Added `HEAD` request to check the database exists before proceeding to backup.

## How

Improved the error reporting by adding an existence check before proceeding to backup.
Refactored so the previously existing restore existence check
can be shared by backup and restore.
Exists function callback function now contains the backup/restore process.
Moved DB does not exist error code to common and updated messages according to case.
Moved the stream error handler registration earlier because of new callback.
Changed the internal backup/restore method signatures to take the db client object instead of URL.

Updated CHANGES.md.

## Testing

Added test for existence check failure on backup.
Updated mocks with existence checks.
Fixed shallow backup tests to match new signatures.

## Issues

Fixes #212

### DIFF
--- a/app.js
+++ b/app.js
@@ -19,13 +19,14 @@
  * @see module:couchbackup
  */
 
-const restoreInternal = require('./includes/restore.js');
-const backupShallow = require('./includes/shallowbackup.js');
 const backupFull = require('./includes/backup.js');
 const defaults = require('./includes/config.js').apiDefaults();
-const events = require('events');
-const debug = require('debug')('couchbackup:app');
 const error = require('./includes/error.js');
+const request = require('./includes/request.js');
+const restoreInternal = require('./includes/restore.js');
+const backupShallow = require('./includes/shallowbackup.js');
+const debug = require('debug')('couchbackup:app');
+const events = require('events');
 const fs = require('fs');
 const legacyUrl = require('url');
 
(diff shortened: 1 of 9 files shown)


app.js
```diff
@@ -136,6 +137,32 @@ function addEventListener(indicator, emitter, event, f) {
...
 
+/*
+  Check couchDbUrl is a valid database URL.
+  @param {string} couchDbUrl - Database URL
+  @param {function(err)} callback - error is undefined if DB exists
+*/
+function exists(db, isBackup, callback) {
+  db.head('', function(err) {
+    err = error.convertResponseError(err, function(err) {
+      if (err && err.statusCode === 404) {
```
tomblench: Do we need to consider 401/403 cases or will these be handled after this check?

app.js
```diff
@@ -136,6 +137,32 @@ function addEventListener(indicator, emitter, event, f) {
...
 }
 
+/*
+  Check couchDbUrl is a valid database URL.
+  @param {string} couchDbUrl - Database URL
+  @param {function(err)} callback - error is undefined if DB exists
+*/
+function exists(db, isBackup, callback) {
+  db.head('', function(err) {
+    err = error.convertResponseError(err, function(err) {
```
tomblench: It seems to me that this method would be cleaner if it just called `callback` with `true` or `false`, then the callee can construct the appropriate error message and you don't need the `isBackup` flag.

However I'm also happy for you to leave as-is, since calling a callback with an error object is probably more nodejs-idiomatic

app.js
ricellis: Good q. The default error factory in the else handles those. The tests that previously asserted 401/403 for _bulk_get have been updated accordingly for the `HEAD` check
i.e. https://github.com/cloudant/couchbackup/blob/212-head-db/test/fatalerrors.js#L128

app.js
ricellis: I see what you mean, the problem is slightly around the range of errors that can be received and constructing the error message with extra information. I didn't really want to have two errorFactory functions that effectively did the same 404 check, but just created a slightly different message. I'll see if I can think of a better way.

ricellis: I think part of the problem here is the naming `exists` makes it sound very simple, but it's more a case of `tryToReachTheDbAndCallbackWithAnErrorIfThereIsSomeProblemAndWeCannotProceed` - that seems a bit longwinded so I've tried `proceedIfDbValid`.
I've also moved the backup/restore message append into the respective functions to avoid the boolean passing.
38051ce5cedc7b2b0e49284143e50fcd301654e4 WDYT?

app.js
```diff
@@ -136,6 +137,32 @@ function addEventListener(indicator, emitter, event, f) {
...
+  visibility. Callback with a fatal error if there is a problem with the DB.
+  @param {string} db - database object
+  @param {function(err)} callback - error is undefined if DB exists
+*/
+function proceedIfDbValid(db, callback) {
+  db.head('', function(err) {
+    err = error.convertResponseError(err, function(err) {
+      if (err && err.statusCode === 404) {
+        // Override the error type and mesasge for the DB not found case
+        var msg = `Database ${db.config.url}/${db.config.db} does not exist. ` +
```
smithsz: We're possibly leaking creds here. Maybe do `db.config.url.replace(/\/\/.+@/g, '//****:****@')`?

ricellis: The message is identical to what it was previously, but it's a fair point, I'll sort it out.

ricellis: Done in 9f6a347a0fbd9290cac47cb82fa8f4e080ba9188

PR #265, opened 2018-12-12
262 enhanced validation

### LINKED ISSUE #262: Backup PARALLELISM
Continuation of #214 

```
ubuntu@cavo_admin:/home/backup$ couchbackup --version
2.3.1
```

```bash
#!/bin/bash

# Couchbackup perfomance options
BUFFERSIZE=10000
MODE="shallow"
PARALLELISM=10

/usr/local/bin/couchbackup --url $URL --buffer-size $BUFFERSIZE --parallelism $PARALLELISM --mode $MODE --log $BASE/$DIR/$DATABASE.log --db $DATABASE > $BASE/$DIR/$DATABASE.json
```

I upgraded the machine I use for backups from 2 procs to 8. It also has a 1 Gbps interface. I have the backups set to:

PARALLELLISM: 10
BUFFERSIZE: 10000

Buffer size is definitely working. I can observe 10000 document for each written batch id:

screen shot 2018-11-29 at 7 36 25 pm

Parallelism just seems broken. I have it set for 10 processes but there are only 9. Additionally, only one of them is doing any serious work as evidenced from htop:

proc

ricellis:
## Checklist

## Description

Added validation for options not valid when using `shallow` mode.

Fixes #262 

## Approach

Changed the way defaults are handled to allow for the invalid default values to be removed in the shallow case.
1. The API defaults were already applied _after_ the arg validation so logging warnings for invalid shallow mode options in that API case was straightforward as part of the validation.
1. For the CLI there were a few problems:
    1. The help messages were ugly in some cases including two copies of the default or showing an `undefined` default.
        * Removed all commander defaults, since we were applying them separately anyway. This fixes the help text and avoids the defaults being present in the `program` variable.
        * Added an `undefined` check when creating default option help text.
    1. The CLI defaults were merged into the `commander` options. This made it difficult to stop the defaults being applied only for `shallow` mode.
        * Changed the approach to merge defaults, environment variable options and CLI provided options after parsing the args.
        * Preserves the order of precedence (i.e. CLI > env var > default)
        * Removed `log` `resume` and `parallelism` from the defaults before merging options when shallow mode is specified.

* Corrected URL handling issues exposed by the validation tests being fixed.
* Corrected use of non-legacy URL in s3 example scripts.
* Updated `CHANGES`
* Updated `package-lock`

## Schema & API Changes

- To prevent behaviour change the validation only outputs a warning rather than terminating the program with an error.

## Security and Privacy

- "No change"

## Testing

- Modified existing `#unit Validate arguments` tests because they were not asynchronous and the assertions were not guaranteed to run before the callback. In fact resolving this exposed a couple of code issues introduced by the removal of the legacy URL usage (which had merged to master, but had not been released yet).
    * Refactored tests to be async to ensure assertions are actually performed.
- Added new tests that capture stderr and check for the warning message:
    - warns for log arg in shallow mode
    - warns for resume arg in shallow mode
    - warns for parallism arg in shallow mode

## Monitoring and Logging

- Added log warning messages for invalid options in `shallow` mode.

### DIFF
--- a/app.js
+++ b/app.js
@@ -118,6 +118,18 @@ function validateArgs(url, opts, cb) {
     return;
   }
 
+  // Perform validation of invalid options for shallow mode and WARN
+  // We don't error for backwards compatibility with scripts that may have been
+  // written passing complete sets of options through
+  if (opts && opts.mode === 'shallow') {
+    if (opts.log || opts.resume) {
+      console.warn('WARNING: the options "log" and "resume" are invalid when using shallow mode.');
+    }
+    if (opts.parallelism) {
+      console.warn('WARNING: the option "parallelism" has no effect when using shallow mode.');
+    }
+  }
+
   if (opts && opts.resume) {
     if (!opts.log) {
       // This is the second place we check for the presence of the log option in conjunction with resume
--- a/app.js
+++ b/app.js
@@ -28,7 +28,7 @@ const backupShallow = require('./includes/shallowbackup.js');
 const debug = require('debug')('couchbackup:app');
 const events = require('events');
 const fs = require('fs');
-const legacyUrl = require('url');
+const URL = require('url').URL;
 
 /**
  * Test for a positive, safe integer.
@@ -95,7 +95,7 @@ function validateArgs(url, opts, cb) {
 
   // Validate URL and ensure no auth if using key
   try {
-    const urlObject = new legacyUrl.URL(url);
+    const urlObject = new URL(url);
     // We require a protocol, host and path (for db), fail if any is missing.
     if (urlObject.protocol !== 'https:' && urlObject.protocol !== 'http:') {
       cb(new error.BackupError('InvalidOption', 'Invalid URL protocol.'));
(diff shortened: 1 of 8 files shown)


app.js
```diff
@@ -105,11 +105,11 @@ function validateArgs(url, opts, cb) {
       cb(new error.BackupError('InvalidOption', 'Invalid URL host.'));
       return;
     }
-    if (!urlObject.pathname) {
+    if (!urlObject.pathname || urlObject.pathname === '/') {
       cb(new error.BackupError('InvalidOption', 'Invalid URL, missing path element (no database).'));
       return;
     }
-    if (opts && opts.iamApiKey && urlObject.auth) {
```
emlaver: Is this `auth` referenced anywhere in the docs that should also be removed?

test/app.js
```diff
@@ -21,97 +21,135 @@ const app = rewire('../app.js');
...
+      done(e);
+    }
+  };
+}
+
+function assertNoError(done) {
+  return function(err, data) {
+    try {
+      assert(err === null, 'There should be no error message.');
+      done();
```
emlaver: What about checking the `data` if not null?

app.js
ricellis: No, the `urlObject` is completely internal, but I searched anyway and can see there is a related `auth` issue from the earlier legacy URL changes in the example scripts. I guess I'll tidy that up here too.

test/app.js
ricellis: I initially tried, but on closer inspection the data is always `null` at the validation stage (the callback only gets called with data when the backup completes). So it's not a valid check.

app.js
ricellis: Done in 44227bf

examples/s3-backup-stream.js
```diff
@@ -169,8 +169,7 @@ function backupToS3(sourceUrl, s3Client, s3Bucket, s3Key, shallow) {
  */
 function s(originalUrl) {
   var parts = new url.URL(originalUrl);
-  delete parts.auth;
-  return url.format(parts);
```
emlaver: Copyright update for this file

ricellis: Done in e127bfe20c4af9e7ce223a8788ef9a2c6a6422c5

PR #318, opened 2020-02-28
Nano8.2.0

glynnbird:
## Checklist

## Description

Nano 8.2 (a dependency of nodejs-cloudant) no longer accepts a blank string as a parameter for `db.head`. The couchbackup code was relying on this to detect the presence of a database. I've swapped out this call for `db.info` which is the equivalent of `GET /db` instead. I've updated the test suite too.

See https://github.com/cloudant/nodejs-cloudant/issues/416

## Approach

```js
db.head('', callback)
```

used to work, but Nano is now more strict on empty or missing parameters. This has been swapped out for:

```js
db.info(callback)
```

which performs the same logical function - "does the database exist?".

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

Modified existing tests because HEAD swapped for GET.

## Monitoring and Logging

- "No change"


CHANGES.md
```diff
@@ -1,3 +1,6 @@
+# 2.5.2-SNAPSHOT (2020-02-28)
+- [FIXED] Isssue with compatibility with Nano 8.2.0.
```
ricellis: 1 too many `s` here

app.js
```diff
@@ -160,7 +160,7 @@ function addEventListener(indicator, emitter, event, f) {
   @param {function(err)} callback - error is undefined if DB exists
 */
 function proceedIfDbValid(db, callback) {
-  db.head('', function(err) {
+  db.info(function(err, data) {
```
ricellis: I know it's not a lot of wasted data, but I think we ought to stick with doing the `HEAD`
```suggestion
  db.server.request({ db: db.config.db, method: 'HEAD' }, function(err) {
```

test/fatalerrors.js
```diff
@@ -113,35 +113,35 @@ function restoreHttpError(opts, errorName, errorCode, done) {
     describe('for backup', function() {
       it('should terminate when DB does not exist', function(done) {
         // Simulate existence check
-        nock(url).head('/fakenockdb').reply(404, { error: 'not_found', reason: 'missing' });
+        nock(url).get('/fakenockdb').reply(404, { error: 'not_found', reason: 'missing' });
```
ricellis: Obviously if we stick with `HEAD` these test changes can be discarded.

CHANGES.md
glynnbird: ha

PR #365, opened 2021-08-06
357 update cloudant client

### LINKED ISSUE #357: Migrate from `@cloudant/cloudant` to `@ibm-cloud/cloudant`

## Bug Description

Migrate to use [Cloudant's new Node SDK](https://github.com/IBM/cloudant-node-sdk/) `@ibm-cloud/cloudant`

Dependency change, no demonstration needed.

Use the latest Cloudant SDK.

### 3. What actually happened

`@cloudant/cloudant` library uses the deprecated `request` library and will itself be deprecated.

## Environment details
    > `<= 2.6.1`
- Node.js version
    > Any

ricellis:
## Checklist

## Description

Replace `@cloudant/cloudant` with `@ibm-cloud/cloudant`

Fixes #357

## Approach

* Replace `@cloudant/cloudant` package dependency with:
    * `@ibm-cloud/cloudant`
    * `tough-cookie` (for session cookie support)
    * `retry-axios` (for retry support)

* Make `request` client return an object containing `service` client and `db` for database name.
* Update service requests to use new calls and models.
* Use promise responses and update response handling where needed.
* Fix broken termination error handling.

## Schema & API Changes

- No API changes.
- Minor changes in non-API:
    -  retry timings
    -  error messages - this is the biggest user facing change here. Error codes remain the same, but the request URL is no longer trivially available to form part of `couchbackup`'s error message.

## Security and Privacy

- "No change"

## Testing

- Modified existing tests because:
    * Mocks needed updating to match new request methods.
    * Some mock responses needed functions to ensure the status message was present.
    * Message assertion updates.

## Monitoring and Logging

- "No change"

### DIFF
--- a/includes/spoolchanges.js
+++ b/includes/spoolchanges.js
@@ -1,4 +1,4 @@
-// Copyright © 2017, 2018 IBM Corp. All rights reserved.
+// Copyright © 2017, 2021 IBM Corp. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -14,10 +14,30 @@
 'use strict';
 
 const fs = require('fs');
+const stream = require('stream');
 const liner = require('./liner.js');
 const change = require('./change.js');
 const error = require('./error.js');
 const debug = require('debug')('couchbackup:spoolchanges');
+
+// Class for streaming _changes error responses into
+// In general the response is a small error/reason JSON object
+// so it is OK to have this in memory.
+class ResponseWriteable extends stream.Writable {
+  constructor(options) {
+    super(options);
+    this.data = [];
+  }
+
+  _write(chunk, encoding, callback) {
+    this.data.push(chunk);
+    callback();
+  }
+
+  stringBody() {
+    return Buffer.concat(this.data).toString();
+  }
+}
 
 /**
  * Write log file for all changes from a database, ready for downloading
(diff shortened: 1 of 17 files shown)


includes/spoolchanges.js
```diff
@@ -62,34 +63,55 @@ module.exports = function(db, log, bufferSize, ee, callback) {
...
+          callback(new error.BackupError('SpoolChangesError', 'Changes request terminated before last_seq was sent'));
+        } else {
+          debug('finished streaming database changes');
+          logStream.end(':changes_complete ' + lastSeq + '\n', 'utf8', callback);
+        }
+      });
+  }).catch(err => {
+    if (err.code && err.code >= 400) {
+      if (err.body) {
+        class ResponseWriteable extends stream.Writable {
```
mojito317, line 89: A comment should be good here, what this is for. For _bulk_docs there was one previously:
```
    // Class for streaming _bulk_docs responses into
    // In general the response is [] or a small error/reason JSON object
    // so it is OK to have this in memory.
```

test/hooks.js
```diff
@@ -15,8 +15,18 @@
 /* global beforeEach afterEach */
 'use strict';
 
-const cloudant = require('@cloudant/cloudant')(
-  (process.env.COUCH_BACKEND_URL) ? { url: process.env.COUCH_BACKEND_URL } : 'https://no-couch-backend-url-set.test');
+const { CloudantV1 } = require('@ibm-cloud/cloudant');
+const url = new URL((process.env.COUCH_BACKEND_URL) ? process.env.COUCH_BACKEND_URL : 'https://no-couch-backend-url-set.test');
+const { BasicAuthenticator, NoAuthAuthenticator } = require('ibm-cloud-sdk-core');
+const authenticator = (url.username) ? new BasicAuthenticator({ username: url.username, password: url.password }) : new NoAuthAuthenticator();
```
mojito317, line 21: Handling missing username/password would be a nice to have.

test/request.js
```diff
@@ -34,46 +34,58 @@ describe('#unit Check request response error callback', function() {
...
+      });
 
-    db.get('bad', function(err) {
+    db.service.getDocument({ db: db.db, docId: 'bad' }).then(response => {
+      done(new Error('Successful response when error expected.'));
+    }).catch(err => {
       err = error.convertResponseError(err);
       assert.strictEqual(err.name, 'HTTPFatalError');
-      assert.strictEqual(err.message, `500 : GET ${url}/bad - Error: foo, Reason: bar`);
+      assert.strictEqual(err.message, '500 foo - Error: foo, Reason: bar');
```
mojito317, line 61: Are the tests checking somewhere there were 3 attempts?

mojito317 (approved):
LGTM! Just some minor comments.

test/request.js
ricellis: yes, the `couch.isDone()` on the next line validates that all the queued responses were returned (i.e. `times(3)` was satisfied in this case).
If a fourth request happens nock will throw an error that there was no matching request.

test/hooks.js
ricellis: I'm not sure I follow - if the `username` is missing we use `NoAuthAuthenticator` - which is a valid use case for our mocks (which don't mock the session handling aspects) or indeed if one was testing against e.g. an admin party CouchDB.
In the case that the `username` is supplied but not the `password` then if a password is required by the server the test will fail anyway wtih a `401`.
Do you mean to make an assumption that a `password` is always needed if a `username` is supplied and assert it separately?

mojito317: > In the case that the `username` is supplied but not the `password` then if a password is required by the server the test will fail anyway wtih a `401`.
> Do you mean to make an assumption that a `password` is always needed if a `username` is supplied and assert it separately?

Yeah, but not necessary, because it will fail on the server-side (that's why I wrote _a nice to have_).

includes/spoolchanges.js
ricellis: Done in 290848be044e9fffd74297fdb1234c10b12b723c - I also pulled the class out of the function for a readability improvement.

emlaver (approved):
Looks great!

PR #381, opened 2021-11-01
Add user-agent in interceptor

### LINKED ISSUE #380: Incorrect user-agent
## Bug Description

When debugging #379 it was discovered that the `user-agent` header is reporting `cloudant-node-sdk` instead of `cloudant-couchbackup`.

Check the `user-agent` header on any request.

The `user-agent` should report `cloudant-couchbackup` and some version metadata.

### 3. What actually happened

The `user-agent` is `cloudant-node-sdk` + version medata.

Although we set the header here: https://github.com/cloudant/couchbackup/blob/68c802b641673d0a29f560970fb70ce90e18cf56/includes/request.js#L134

The SDK sets the `user-agent` header later than that, so overrides the value provided by couchbackup. We need to set the header even later to make sure it is correct to assist with debugging.

## Environment details
    > 2.7.0
- Node.js version
    > 14.18.1

ricellis:
## Checklist

## Description

Set the `user-agent` header to the correct value.

Fixes #380

## Approach

Set the `user-agent` header in an interceptor to avoid the SDK overriding it with its own value.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- Added new tests to assert `user-agent` header is correct.

## Monitoring and Logging

- Request header updated so easier to log/debug requests originating from couchbackup.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,6 @@
+# Unreleased
+- [FIXED] Corrected `user-agent` header on requests.
+
 # 2.7.0 (2021-09-14)
 - [UPGRADED] Cloudant client dependency from `@cloudant/cloudant` to `@ibm-cloud/cloudant`.
 
--- a/includes/request.js
+++ b/includes/request.js
@@ -131,7 +131,6 @@ module.exports = {
     const serviceOpts = {
       authenticator: authenticator,
       timeout: opts.requestTimeout,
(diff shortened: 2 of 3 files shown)


emlaver (approved):
Looks good

mojito317 (approved):

PR #384, opened 2021-11-04
Fix restore of old version shallow backups

### LINKED ISSUE #379: Version 2.7.0 cannot restore shallow backups from before 2.4.2

1. Make a shallow backup with a version `<2.4.2`
2. Try to restore the backup with version `2.7.0`

The restore should succeed.

Although we haven't made a statement anywhere about mixing versions of backup and restore I think there is no real reason this shouldn't work and I could easily imagine a backup project running on a cron with an old version and not attempting to do a restore until it was really needed and grabbing the lasest version at the time. The failure mode was introduced in `2.7.0`.

### 3. What actually happened

The restore fails with: `ERROR: Error writing batch.` This happens because:
https://github.com/cloudant/couchbackup/blob/68c802b641673d0a29f560970fb70ce90e18cf56/includes/writer.js#L42
expects the empty array response of `new_edits: false` mode.
However, https://github.com/cloudant/couchbackup/blob/68c802b641673d0a29f560970fb70ce90e18cf56/includes/writer.js#L33
still toggles `new_edits:true` (note: the absense of the parameter gives the default `true`) based on whether there is any `_rev` info in the payload. It is worth noting that all backups produced after 2.4.2 include `_rev` info, but shallow backups created prior to `2.4.2` did not include `_rev` info.

We should either:
Toggle the response checking behaviour based on the `new_edits` parameter
or
Remove the toggle and document something about backup/restore version compatibility

Also the error message here at https://github.com/cloudant/couchbackup/blob/68c802b641673d0a29f560970fb70ce90e18cf56/includes/writer.js#L43  could probably do with a bit more information like indicating if the `result` was missing or non-empty - and potentially in the `new_edits:false` case extracting error responses from the set. (We might also want to consider checking for non `"ok":true` entries in the `new_edits:true` case, but I'm concerned that might have a significant impact on restore performance).

## Environment details
    > 2.7.0
- Node.js version
    > 12.22.7

ricellis:
## Checklist

## Description

Correct batch error check on restore of shallow backups created with versions <2.4.2

Fixes #379 

## Approach

Correct batch error check on restore wrriterr batch response to honour the `new_edits` toggle in the payload that could be triggered on by shallow backups that don't include rev info (i.e. those created with versions earlier than `2.4.2`).

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- Added new tests:
    - `should restore shallow backups without rev info successfully` with fixture of a revisionless shallow backup

## Monitoring and Logging

- Augment the error message to make debugging slightly easier in the case of batch write errors.


includes/writer.js
```diff
@@ -32,15 +32,17 @@ module.exports = function(db, bufferSize, parallelism, ee) {
...
 
     if (!didError) {
       db.service.postBulkDocs({
         db: db.db,
         bulkDocs: payload
       }).then(response => {
-        if (!response.result || response.result.length > 0) {
-          throw new Error('Error writing batch.');
+        if (!response.result || (payload.new_edits === false && response.result.length > 0)) {
+          throw new Error(`Error writing batch with new_edits:${payload.new_edits !== false}` +
```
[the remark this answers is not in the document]
ricellis: Sure added in 1edf6cf76222d39af34a8ecc7227810f706b723b

includes/writer.js
```diff
@@ -32,15 +32,17 @@ module.exports = function(db, bufferSize, parallelism, ee) {
...
     }
 
     if (!didError) {
       db.service.postBulkDocs({
         db: db.db,
         bulkDocs: payload
       }).then(response => {
-        if (!response.result || response.result.length > 0) {
-          throw new Error('Error writing batch.');
+        if (!response.result || (payload.new_edits === false && response.result.length > 0)) {
```
mojito317, line 43: IMO a test case when payload does not have `new_edits` would fail here. Can this happen in practice?
E.g. the payload arrives without revs AND without the `new_edits` param?

You wrote in the issue description:
> It is worth noting that all backups produced after `2.4.2` include `_rev` info, but shallow backups created prior to `2.4.2` did not include `_rev` info.

ricellis: >a test case when payload does not have new_edits would fail here

fail the `if` check, or return the error?

>Can this happen in practice?
>E.g. the payload arrives without revs AND without the new_edits param?

yes, this is exactly what happens in a restore of a <2.4.2 shallow backup

The setting of the `new_edits` property is governed by the presence of `_rev` in the doc content of the backup (it's [line 33/34](https://github.com/cloudant/couchbackup/pull/384/files/1edf6cf76222d39af34a8ecc7227810f706b723b#diff-3b76bddc4e460cc434375132ce2fbdd645bc47451c2d93512bb00df275e60296R34))

This means that for backups without `_rev` info, `new_edits` is `undefined` and the restore has `new_edits=true` behaviour on the request.

That is the reason I used the check `payload.new_edits === false` and not `payload.new_edits` because the `undefined` `new_edits` state is _falsy_, but does not correspond to `new_edits=false` behaviour and in the `new_edits=true` (i.e. `undefined`) case we don't want to error if the array has content.

This is the behaviour I wanted:
| `payload.new_edits` | `new_edits` on request | should check array length=0 |
| --- | --- | --- |
| `false` | `false` | yes, expect empty array |
| `true` | `true` | no, expect array content (note we don't explcitly set `true`) |
| `undefined` | `true` | no, expect array content |

Do you think the code delivers that? Did I miss something?

mojito317: > fail the `if` check, or return the error?

Meanwhile, I realized Node has the type check (triple =s). I think I spent too much time in a Python environment lately huh? So basically if `payload.new_edits` is `undefined`, then `payload.new_edits === false` will be false, gotcha.

emlaver (approved):
looks good, thanks for adding the test

mojito317 (approved):
lgtm!

includes/writer.js
ricellis: np, thanks for being thorough better to double-check now than break later.

PR #409, opened 2021-11-15
Extend preflight check in a database restore function

### LINKED ISSUE #376: Add a preflight check for restore function to make sure that target database is new
Couchbackup could benefit from a perflight check for restore functionality to confirm that a target database exists and is new , i.e. it has `doc_count` and `doc_del_count` in a database info set to `0`.

There seems to be a possibility of confusion on the requirements for a target database (ref: #367), so it'd make it a better user experience to have this part to be an explicit run error or at least a warning.

It looks like `includes/writer.js` is the right spot to add this functionality.

eiri:
## Checklist

## Description

This change extends a preflight check in a database restore function to make sure that a target database is a new and empty. It does that by confirming that the target database has 0 actual or deleted docs.

The check adds a new error `DatabaseNotEmpty` with an exit code 13.

Fixes #376 

## Approach

Function `proceedIfDbValid` got split in two version for backup and restore, where restore version was changed to check that number of docs and deleted docs in a target database is zero.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- Added new tests:
    - should terminate on `notEmptyDBErr` when database is not empty
    - should terminate on `notEmptyDBErr` when database is not new
    
- Modified existing restore tests to reflect changed validation approach

## Monitoring and Logging

- "No change"

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,6 +425,8 @@ details them.
 
 ### `couchrestore`
 
+* `13`: restore target database is not new and empty.
+
 ## Note on attachments
 
 TLDR; If you backup a database that contains attachments you will not be able to restore it.
--- a/app.js
+++ b/app.js
@@ -176,7 +176,7 @@ function proceedIfRestoreDbValid(db, callback) {
   db.service.getDatabaseInformation({ db: db.db }).then(response => {
     const { doc_count: docCount, doc_del_count: deletedDocCount } = response.result;
     if (!db.db.startsWith("_") && (docCount !== 0 || deletedDocCount !== 0)) {
-      var notEmptyDBErr = new Error(`Database ${db.url} is not empty.`);
+      var notEmptyDBErr = new Error(`Target database ${db.url} is not empty.`);
       notEmptyDBErr.name = 'DatabaseNotEmpty';
       callback(notEmptyDBErr);
     } else {
--- a/app.js
+++ b/app.js
@@ -175,7 +175,7 @@ function proceedIfBackupDbValid(db, callback) {
 function proceedIfRestoreDbValid(db, callback) {
   db.service.getDatabaseInformation({ db: db.db }).then(response => {
     const { doc_count: docCount, doc_del_count: deletedDocCount } = response.result;
-    if (docCount !== 0 || deletedDocCount !== 0) {
+    if (!db.db.startsWith("_") && (docCount !== 0 || deletedDocCount !== 0)) {
       var notEmptyDBErr = new Error(`Database ${db.url} is not empty.`);
       notEmptyDBErr.name = 'DatabaseNotEmpty';
       callback(notEmptyDBErr);
--- a/app.js
+++ b/app.js
@@ -199,7 +199,7 @@ function proceedIfRestoreDbValid(db, callback) {
 */
 function parseIfDbValidResponseError(db, err) {
   if (err && err.status === 404) {
-    // Override the error type and mesasge for the DB not found case
+    // Override the error type and message for the DB not found case
     var msg = `Database ${db.url}` +
     `${db.db} does not exist. ` +
     'Check the URL and database name have been specified correctly.';
--- a/app.js
+++ b/app.js
@@ -175,8 +175,11 @@ function proceedIfBackupDbValid(db, callback) {
 function proceedIfRestoreDbValid(db, callback) {
   db.service.getDatabaseInformation({ db: db.db }).then(response => {
     const { doc_count: docCount, doc_del_count: deletedDocCount } = response.result;
+    // The system databases can have a validation ddoc(s) injected in them on creation.
+    // This sets the doc count off, so we just complitely exclude the system databases from this check.
+    // The assumption here is that users restoring system databases know what they are doing.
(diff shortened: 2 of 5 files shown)


app.js
```diff
@@ -154,31 +154,60 @@ function addEventListener(indicator, emitter, event, f) {
...
+  Check the restore database exists, new and empty and that the credentials used have
+  visibility. Callback with a fatal error if there is a problem with the DB.
+  @param {string} db - database object
+  @param {function(err)} callback - error is undefined if DB exists, new and empty
+*/
+function proceedIfRestoreDbValid(db, callback) {
+  db.service.getDatabaseInformation({ db: db.db }).then(response => {
+    const { doc_count: docCount, doc_del_count: deletedDocCount } = response.result;
+    if (!db.db.startsWith("_") && (docCount !== 0 || deletedDocCount !== 0)) {
+      var notEmptyDBErr = new Error(`Database ${db.url} is not empty.`);
```
emlaver, line 179:
```suggestion
      var notEmptyDBErr = new Error(`Restore database ${db.url} is not empty.`);
```

app.js
```diff
@@ -154,31 +154,60 @@ function addEventListener(indicator, emitter, event, f) {
...
+      notEmptyDBErr.name = 'DatabaseNotEmpty';
```
emlaver, line 180:
```suggestion
      notEmptyDBErr.name = 'RestoreDatabaseNotEmpty';
```

README.md
```diff
@@ -425,6 +425,8 @@ details them.
 
 ### `couchrestore`
 
+* `13`: restore target database is not new and empty.
```
emlaver, line 428: what about:
```suggestion
* `13`: restore target database must be a new and empty database.
```

app.js
eiri: Then "target database", not "restore",  to match how we call it in docs.

README.md
eiri: This doesn't match the tone of the rest error descriptions though, they are descriptive and not prescriptive, e.g. "10: backup source or restore target database does not exist." not - "10: backup source or restore target database must exist", if you see what I mean.

app.js
emlaver: Ok "target database" sounds good to match the docs.

README.md
emlaver: Yea good point, let's keep what you have then.

app.js
emlaver: Ok no problem, I'm fine with keeping it as-is.  I figured I'd bring it up if we think it's better to have it more specific especially if this function will only be used by restore functionality.

app.js
ricellis: I think the message will need to be augmented here to be `${db.url}${db.db}` similar to the not found case otherwise this is only going to present the account URL and not the specific empty DB (ofc it is a required user supplied arg so it ought to be obvious which DB, but I think a more complete message will help in cases where people are iterating over all_dbs etc).

And yes, the slightly goofy naming made this very easy to overlook since `db.url` is actually the service URL - historically `db` was a database instance, but when we changed the underlying client it became an object holding a ref to the client (`db.service`), service url (`db.url`) and db name (`db.db`).

app.js
```diff
@@ -154,31 +154,60 @@ function addEventListener(indicator, emitter, event, f) {
...
+/*
+  Check that the restore database exists, is new and is empty. Also verify that the credentials used have
+  visibility. Callback with a fatal error if there is a problem with the DB.
+  @param {string} db - database object
+  @param {function(err)} callback - error is undefined if DB exists, new and empty
+*/
+function proceedIfRestoreDbValid(db, callback) {
+  db.service.getDatabaseInformation({ db: db.db }).then(response => {
+    const { doc_count: docCount, doc_del_count: deletedDocCount } = response.result;
+    if (!db.db.startsWith("_") && (docCount !== 0 || deletedDocCount !== 0)) {
```
ricellis, line 178: Might be worth a comment here about why we exclude the `_` prefixed DBs from the check

ricellis (commented):
A couple of minor nits.

eiri:
@ricellis @emlaver sorry for the force push, I rebased my dev branch over master's HEAD out of habit instead of merging it.

ricellis (approved):
+1

app.js
```diff
@@ -154,31 +154,63 @@ function addEventListener(indicator, emitter, event, f) {
...
 
+/*
+  Convert the database validation response error to a special DatabaseNotFound error
+  in case the database is missing. Otherwise delegate to the default error factory.
+  @param {object} db - database object
+  @param {object} err - HTTP response error
+*/
+function parseIfDbValidResponseError(db, err) {
+  if (err && err.status === 404) {
+    // Override the error type and mesasge for the DB not found case
```
mojito317, line 202: When I skim through this PR I found a typo:
```suggestion
    // Override the error type and message for the DB not found case
```

eiri: Ah, thank you! This is a prior art, not my change, so haven't noticed that 😂

emlaver (approved):

eiri:
All right, @ricellis @emlaver @mojito317 thank you for the review, I'm merging this.

PR #412, opened 2021-11-16
Split changes spooling

### LINKED ISSUE #101: Handle `last_seq` for Cloudant maintenance events
Cloudant terminates the changes feed, sending a `last_seq`, if the node serving your request is placed in maintenance mode (see https://github.com/apache/couchdb-fabric/blob/master/src/fabric_view_changes.erl#L94). The expectation here is that the client will retry the request using the `last_seq` provided in the response.

The current backup code assumes that the changes feed has finished if a `last_seq` is returned. We need to be sure that this is indeed the case by validating the number of changes we've received with the database `update_seq` (or similar).

### LINKED ISSUE #408: Cannot complete spooling changes for some DBs
## Bug Description

The spooling changes phase attempts to complete in a single request. For DBs with large numbers of changes (more typically large numbers of documents) this may not be possible before the connection is terminated. The lifetime of the connection may be determined by CouchDB settings for changes feeds or by limits imposed by proxies.

Note it is possible to workaround this by using shallow mode backups (but obviously they do not contain equivalent data to full mode).

Perform full backup on a database with >~36M documents (there is no exact number; it is really the _time_ to spool changes that is the trigger, but they are closely linked).

Backup to complete successfully.

### 3. What actually happened

Backup fails because the changes spooling is unable to complete before the connection is teminated resulting in an:
`Error: unexpected end of file`

## Environment details
    > All full backups
- Node.js version
    > Any

ricellis:
## Checklist

## Description
Split changes spooling into multiple requests.

Fixes #408 
I believe this will also close #101 (as `last_seq` responses will now be handled in the loop) and #100 (since the built-in retries will now apply per batch request for changes and not only for a single request i.e. retries of any given request will be from that `last_seq` not the beginning of the feed). 

## Approach

Apply a `limit` to changes requests and iterate with `since` and `last_seq`.

_Reviewer notes_: what is an appropriate `limit` to use? I've chosen 100k as a starting bid.

## Schema & API Changes

- No API change, underlying change in behaviour should be transparent to users.

## Security and Privacy

- "No change"

## Testing

- Added new tests to cover collection of changes over multiple requests:
    - `should keep collecting changes` - to validate multiple requests are made
    - `should keep collecting sparse changes` - to validate thousands of requests don't cause issues
    - `Longer spool changes checks -#slow should keep collecting changes (25M)` - checks 25M changes
    - `Longer spool changes checks  - #slower should keep collecting changes (500M)` - checks 500M full changes

## Monitoring and Logging

- Added debug line for each request with `since` value

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 - [IMPROVED] Added quiet option to backup and restore to suppress batch messages.
 - [IMPROVED] Added a preflight check for restore function to make sure that a target database is new and empty.
 - [IMPROVED] Added handling for errors reading log file.
+- [IMPROVED] Split changes spooling to improve reliability on databases with
+  millions of documents.
 
 # 2.7.0 (2021-09-14)
 - [UPGRADED] Cloudant client dependency from `@cloudant/cloudant` to `@ibm-cloud/cloudant`.
--- a/includes/spoolchanges.js
+++ b/includes/spoolchanges.js
@@ -34,6 +34,9 @@ module.exports = function(db, log, bufferSize, ee, callback) {
   let batch = 0;
   let lastSeq = null;
   const logStream = fs.createWriteStream(log);
+  let pending = 0;
+  // The number of changes to fetch per request
+  const limit = 100000;
 
   // send documents ids to the queue in batches of bufferSize + the last batch
   const processBuffer = function(lastOne) {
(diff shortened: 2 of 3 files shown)


vmatyusGitHub (approved):
I am okay with the 100k limit number.

emlaver (approved):
Looks good! I'm also fine with 100k limit.

eiri (approved):
legit. tested on 1.2mil local database

```
$ curl -u xxx:xxx  https://localhost:5984/koi | jq .doc_count
1200000

$ npx ./bin/couchbackup.bin.js --url https://xxx:xxx@localhost:5984 --db koi > koi.json
...
couchbackup:backup:batch Written batch ID: 2399 Total document revisions written: 1200000 Time: 873.703 +54ms
couchbackup:backup Finished - Total document revisions written: 1200000 +15m

$ cat koi.json | npx ./bin/couchrestore.bin.js --url https://xxx:xxx@localhost:5984 --db koi2
...
couchbackup:restore:batch restored 1200000 +15s
couchbackup:restore finished { total: 1200000 } +0ms

$ curl -u xxx:xxx  https://localhost:5984/koi2 | jq .doc_count
1200000
```

PR #468, opened 2022-04-21
Properly urlescape password in tests hook and Jenkins pipeline

eiri:
## Checklist

- [ ] Updated the change log file (`CHANGES.md`|`CHANGELOG.md`) _or_ test/build only changes

## Description

This PR properly encodes password in Jenkins pipeline [as required by our own README](https://github.com/cloudant/couchbackup#the-url) and decodes it in test's hook, since node's `new URL` doesn't do that. This solves an issue with test password that might contain URL components.

Also this PR bumps `eslint` and `eslint-config-standard` to unlock its cross deps, that break `npm ci`

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -63,6 +63,7 @@ def setupNodeAndTest(version, filter='', testSuite='test') {
               // For the IAM tests we want to run the normal 'test' suite, but we
               // want to keep the report named 'test-iam'
               def testRun = (testSuite != 'test-iam') ? testSuite : 'test'
+              def dbPassword = java.net.URLEncoder.encode(DB_PASSWORD, "UTF-8")
 
               // Actions:
               //  1. Load NVM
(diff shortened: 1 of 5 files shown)


Jenkinsfile
```diff
@@ -77,8 +78,10 @@ def setupNodeAndTest(version, filter='', testSuite='test') {
                 npm install mocha-jenkins-reporter --save-dev
                 curl -O -u "\${ARTIFACTORY_USER}:\${ARTIFACTORY_PW}" "https://na.artifactory.swg-devops.com/artifactory/cloudant-sdks-maven-local/com/ibm/cloudant/${env.DBCOMPARE_NAME}/${env.DBCOMPARE_VERSION}/${env.DBCOMPARE_NAME}-${env.DBCOMPARE_VERSION}.zip"
                 unzip ${env.DBCOMPARE_NAME}-${env.DBCOMPARE_VERSION}.zip
-                export COUCH_BACKEND_URL="https://\${DB_USER}:\${DB_PASSWORD}@\${SDKS_TEST_SERVER_HOST}"
+                set +x
+                export COUCH_BACKEND_URL="https://\${DB_USER}:${dbPassword}@\${SDKS_TEST_SERVER_HOST}"
```
vmatyusGitHub, line 82: Have you tried to use: https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#interpolation-of-sensitive-environment-variables?
```suggestion
                export COUCH_BACKEND_URL='https://\${DB_USER}:${dbPassword}@\${SDKS_TEST_SERVER_HOST}'
```
Based on my local test, this will result:
`export COUCH_BACKEND_URL=https://${DB_USER}:${DB_PASSWORD}@${SDKS_TEST_SERVER_HOST}`

vmatyusGitHub:
> since node's new URL doesn't do that.

The opposite, is true `new URL` does encoding in `hook.js`.
But as I see the DB comparison fails even after password decoding.
Might this is the problem, an encoded url should be stored in env var to be a match with the backed-up DB.

Here is a small node snippet:
```
> url = "http://user:p@ss@localhost"
'http://user:p@ss@localhost'
> new URL(url)
URL {
  href: 'http://user:p%40ss@localhost/',
  origin: 'http://localhost',
  protocol: 'http:',
  username: 'user',
  password: 'p%40ss',
  host: 'localhost',
  hostname: 'localhost',
  port: '',
  pathname: '/',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```

Jenkinsfile
vmatyusGitHub: `sh'''` syntax is used during git config in sdks.

eiri: no, this doesn't work on two levels: 1) this doesn't mask password 2) it does direct interpolation of vars, so we are not getting _value_ of `${DB_USER}`, but literal name of variable, e.g. it looks like `COUCH_BACKEND_URL=https://${DB_USER}:xxxx@${SDKS_TEST_SERVER_HOST}` which obviously doesn't work

eiri:
@vmatyusGitHub yes, this is exactly what I said, node's doesn't do **decoding** of urlencoded password, so this need to be done in `hook`

vmatyusGitHub (approved):

PR #530, opened 2022-11-21
chore: pull npm dependencies from artifactory

tomblench:


Jenkinsfile
```diff
@@ -94,11 +98,42 @@ def setupNodeAndTest(version, filter='', testSuite='test') {
...
+
+// url of registry for public uploads
+def getRegistryPublic() {
+    return "https://registry.npmjs.org/"
+}
+
+// url of registry for artifactory up
+def getRegistryArtifactoryUp() {
+    return "${Artifactory.server('taas-artifactory-upload').getUrl()}/api/npm/cloudant-sdks-npm-virtual/"
+}
```
ricellis, line 111: Unneeded here.
```suggestion
```

.npmrc-jenkins
```diff
@@ -0,0 +1,13 @@
...
+${NPM_REGISTRY}:email=${NPM_EMAIL}
+${NPM_REGISTRY}:_authToken=${NPM_TOKEN}
+
+; for artifactory virtual
+${ARTIFACTORY_DOWN}:username=${ARTIFACTORY_TOKEN_USR}
+${ARTIFACTORY_DOWN}:email=${ARTIFACTORY_TOKEN_USR}
+${ARTIFACTORY_DOWN}:_authToken=${ARTIFACTORY_TOKEN_PSW}
+${ARTIFACTORY_UP}:username=${ARTIFACTORY_TOKEN_USR}
+${ARTIFACTORY_UP}:email=${ARTIFACTORY_TOKEN_USR}
+${ARTIFACTORY_UP}:_authToken=${ARTIFACTORY_TOKEN_PSW}
```
ricellis, line 13: Unneeded here.
```suggestion
```

Jenkinsfile
```diff
@@ -75,15 +75,19 @@ def setupNodeAndTest(version, filter='', testSuite='test') {
...
-                unzip ${env.DBCOMPARE_NAME}-${env.DBCOMPARE_VERSION}.zip
-                set +x
-                export COUCH_BACKEND_URL="https://\${DB_USER}:${dbPassword}@\${SDKS_TEST_SERVER_HOST}"
-                export COUCH_URL="${(testSuite == 'toxytests/toxy') ? 'http://localhost:3000' : ((testSuite == 'test-iam') ? '${SDKS_TEST_SERVER_URL}' : '${COUCH_BACKEND_URL}')}"
-                set -x
-                ./node_modules/mocha/bin/mocha.js --reporter mocha-jenkins-reporter --reporter-options junit_report_path=./test/test-results.xml,junit_report_stack=true,junit_report_name=${testSuite} ${filter} ${testRun}
-              """
+                """
+              withNpmEnv('ARTIFACTORY_DOWN', registryArtifactoryDown) {
+                sh """
```
ricellis, line 80: Historically the NVM commands and the rest of the `sh` script had to be contained in the same `sh` step because the NVM setting didn't persist across multiple `sh` invocations. I can't see what version the tests are actally using in the build logs so we'll have to dig a bit deeper. I suspect this may need to be refactored to have the `withNpmEnv` enclose an entire single `sh` step, but I see that caused some other problems before.

Since it looked like the problem happened during install maybe we can resolve by doing the `nvm install` first and then `nvm use` within the same sh step as the test invocation. Otherwise we might need to check the nvm docs and see if we can flag off whatever it is doing.

ricellis (changes requested):

Jenkinsfile
```diff
@@ -71,19 +71,25 @@ def setupNodeAndTest(version, filter='', testSuite='test') {
...
-                unzip ${env.DBCOMPARE_NAME}-${env.DBCOMPARE_VERSION}.zip
-                set +x
-                export COUCH_BACKEND_URL="https://\${DB_USER}:${dbPassword}@\${SDKS_TEST_SERVER_HOST}"
-                export COUCH_URL="${(testSuite == 'toxytests/toxy') ? 'http://localhost:3000' : ((testSuite == 'test-iam') ? '${SDKS_TEST_SERVER_URL}' : '${COUCH_BACKEND_URL}')}"
-                set -x
-                ./node_modules/mocha/bin/mocha.js --reporter mocha-jenkins-reporter --reporter-options junit_report_path=./test/test-results.xml,junit_report_stack=true,junit_report_name=${testSuite} ${filter} ${testRun}
-              """
+              withCredentials([usernamePassword(usernameVariable: 'NPMRC_USER', passwordVariable: 'NPMRC_TOKEN', credentialsId: 'artifactory-id-token')]) {
+              withEnv(['NPMRC_EMAIL=' + env.NPMRC_USER]) {
+              withNpmEnv(registryArtifactoryDown) {
```
ricellis, line 76: nit: indentation

Jenkinsfile
ricellis: the other up refernce is gone now, but still need to get rid of this one

Jenkinsfile
```diff
@@ -94,11 +100,46 @@ def setupNodeAndTest(version, filter='', testSuite='test') {
...
+}
+
 stage('Build') {
   // Checkout, build
   node('sdks-backup-executor') {
     checkout scm
-    sh 'npm ci'
+    withCredentials([usernamePassword(usernameVariable: 'NPMRC_USER', passwordVariable: 'NPMRC_TOKEN', credentialsId: 'artifactory-id-token')]) {
+    withEnv(['NPMRC_EMAIL=' + env.NPMRC_USER]) {
+    withNpmEnv(registryArtifactoryDown) {
```
ricellis, line 138: nit indentation

ricellis (approved):

PR #536, opened 2022-12-09
fix file descriptor leak in backup

### LINKED ISSUE #535: File descriptor leak

## Bug Description
There is a file descriptor leak in this library. We found that after calling `backup` there would be a file left open in the `/tmp` folder.

We run our backup process from an AWS lambda that has a hard limit of 1,024 file descriptors.  Since we are using a lambda, the process doesn't exit straight away and gets reused for the next call.  

Each of our projects is a separate database meaning that we slowly have more databases to back up and In August, we started getting error messages like
`EMFILE: too many open files, open '/tmp/couchbackup/couchbackup-qa_xsxxxxxxxxxxxx-xxxxxxxxx_2022-11-15T17:47:52.144Z.log'` 

This issue is cause by the [apiDefaults](https://github.com/cloudant/couchbackup/blob/f44a8a3fe2967cd33e32dd911b699f019bbee1a4/includes/config.js#L27) function.  When you look into the [`tmp.fileSync()`](https://github.com/raszi/node-tmp/blob/da3bc1ac9a7880def4a3129c4f8bfe96a63bf99c/lib/tmp.js#L171) function you can see that it opens a file which is only closed if you set `opts.discardDescriptor` to true. 

Given that the code is only after a tmp file name it might be better to use [`tmpNameSync`](https://github.com/raszi/node-tmp/blob/da3bc1ac9a7880def4a3129c4f8bfe96a63bf99c/lib/tmp.js#L100)

You can see the number of open fd using Lambda Insights.  This is the graph with the current library.
image

After I changed that call to be `tmp.fileSync({ discardDescriptor: true }).name` 
image

This isn't a lambda but you can see the files being left open.  They will be finally closed when the process exists, but AWS Lambda leaves the process running so it can be reused. 
```js
import { exec } from "child_process";
import { join } from "path";
import os from "os";
import fs from "fs";
import { backup } from "@cloudant/couchbackup";

const listFileDescriptor = () => {
  return new Promise((resolve, reject) => {
    exec(`ls -l /proc/${process.pid}/fd`, (err, stdout, stderr) => {
      if (err) {
        reject(err);
      } else {
        const parts = stdout.split("\n");
        // there will be other fd open, but it's only the tmp- ones causing problems
        const tmpFiles = parts.filter((x) => x.includes("/tmp/"));
        resolve(tmpFiles);
      }
    });
  });
};

const databases = [
    "https://user:password@host/database1",
    "https://user:password@host/database2",
  ];

const backupDatabases = async () => {
  const fdStart = await listFileDescriptor();
  console.log(fdStart);

  const tmpFolder = os.tmpdir();
  const tmpFile = join(
    tmpFolder,
    `couchbackup-${new Date().toISOString()}.log`
  );

  await backup(databases[0], fs.createWriteStream("dump1.json"), (err) => {
    if (err) {
      return new Error(err, "CouchBackup process failed");
    }
    console.log("couchbackup to file done;");
  });

  await backup(databases[1], fs.createWriteStream("dump1.json"), (err) => {
    if (err) {
      return new Error(err, "CouchBackup process failed");
    }
    console.log("couchbackup to file done;");
  });

  const fdEnd = await listFileDescriptor();
  console.log(fdEnd);
};

backupDatabases();


```

There should be no temp files when the `backup` call has completed and the number of open fd should not be increasing. 

### 3. What actually happened
There are two temp files left open.
image

## Environment details

    > 2.9.3
- Node.js version
    >14.20.0

emlaver:
## Checklist

## Description

Replace `tmp.fileSync().name` in [apiDefaults](https://github.com/cloudant/couchbackup/blob/f44a8a3fe2967cd33e32dd911b699f019bbee1a4/includes/config.js#L27) with `tmp.tmpNameSync()` to discard descriptor of temporary file during backup.

fixes #535 

## Approach
Update `tmp` call in `includes/config.js`

## Testing
Used this [gist](https://gist.github.com/emlaver/9239ce5e12182e158b00739825a9e89b) to verify that file descriptors are discarded once backup completes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,6 @@
+# Unreleased
+- [FIXED] Discard temporary file descriptor for backup.
+
 # 2.9.5 (2022-11-01)
 - [UPGRADED]  `@ibm-cloud/cloudant` dependency to version `0.3.0`.
 
--- a/includes/config.js
+++ b/includes/config.js
@@ -24,7 +24,7 @@ function apiDefaults() {
     parallelism: 5,
     bufferSize: 500,
     requestTimeout: 120000,
-    log: tmp.fileSync().name,
+    log: tmp.tmpNameSync(),
     resume: false,
     mode: 'full'
   };


eiri:
To be honest I can't see how this fixes the leaked file descriptors, consider that [`fileSync` is using `tmpNameSync` under the hood](https://github.com/raszi/node-tmp/blob/master/lib/tmp.js#L170). Do you have an explanation why this works?

As an alternative I'd suggest to use `tmp.fileSync({ discardDescriptor: true })` instead, as described in https://github.com/raszi/node-tmp/tree/master#controlling-the-descriptor or go an extra mile and use `detachDescriptor` and manage the descriptor closing at backup `done` callback, though that sounds like a hustle.

ricellis:
>Do you have an explanation why this works?

`tmpNameSync` just returns a name, `fileSync` uses that name to open the file. We actually have no need to open the file when processing the config, we open it later by creating a read stream anyway.

eiri:
So this is essentially using `fstat` vs `fopen`? I'd expect `fd` to still be allocated, tbh, so we'll win on memory and probably op speed, but not on `fd` count reduce, but I'm not 100% sure, maybe system takes care of that.

Ok, thanks for the clarification, no harm in using this.

eiri (approved):

ricellis (approved):

PR #558, opened 2023-02-16
Init SonarQube to pipeline

mojito317:
## Checklist

## Description
This change adds SonarQube Analysis to our pipeline.

## Approach

Introduced the `SonarQube analysis` stage.
The analysis will not happen on `dependabot` generated branches.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

## Monitoring and Logging

- "No change"

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,5 @@
 #!groovy
-// Copyright © 2017, 2019 IBM Corp. All rights reserved.
+// Copyright © 2017, 2023 IBM Corp. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -166,6 +166,18 @@ stage('QA') {
   }
   // Run the required axes in parallel
   parallel(axes)
+}
+
+stage('SonarQube analysis') {
+  node('sdks-backup-executor') {
+    unstash name: 'built'
+    if (!(env.BRANCH_NAME).startsWith('dependabot/')) {
+      def scannerHome = tool 'SonarQubeScanner';
+      withSonarQubeEnv(installationName: 'SonarQubeServer') {
+        sh "${scannerHome}/bin/sonar-scanner -X -Dsonar.projectKey=couchbackup -Dsonar.branch.name=${env.BRANCH_NAME}"
+      }
+    }
+  }
 }
 
 // Publish the primary branch


Jenkinsfile
```diff
@@ -1,5 +1,5 @@
 #!groovy
-// Copyright © 2017, 2019 IBM Corp. All rights reserved.
+// Copyright © 2017, 2023 IBM Corp. All rights reserved.
```
pgj, line 2: Should not this say "2017, 2019, 2023" instead?  For example, https://techwhirl.com/updating-copyright-notices/ says _"Why not just change the date to the most recent year and delete the earlier date(s)? Well, first of all, you’d be creating an inaccurate copyright notice. (Not all of the material was first published in the most recent year, right?) And, more importantly, the law will punish you for doing so. According to the U.S. Copyright Act, if “the year date [in the copyright notice] is more than one year later than the year in which publication first occurred, the work is considered to have been published without any notice….” The consequences, though not fatal to copyright protection, can be significant, so it’s important to use care when deciding what date(s) to include in your copyright notice."_

mojito317: In my understanding, most importantly the year should be there when it first occurred in publication. In this case, it was 2017, which I have not touched. So the work is considered to have been published in 2017, while its last modification was in 2023. I know a `-` would be a better way of connecting the two years, but using `,` is kind of our convention. Several examples can be found in this repo with this. To take an example the [backup.js](https://github.com/cloudant/couchbackup/blob/main/includes/backup.js) file has `Copyright © 2017, 2021 IBM Corp. All rights reserved.` while checking the commit history it was touched in 2017, 2018, 2019, and 2021.

pgj: Well, it might easily happen that the cited copyright notice is wrong too.  The `-` (dash) symbol should be used to denote a range of consecutive years.  If there were gaps, comma-separated ranges may also work, i.e. 2017-2019, 2021 for `backup.js`.  But I am not a lawyer, so this has been a humble observation on my side.

ricellis (approved):

eiri (approved):

Jenkinsfile
eiri: This is correct format for IBM copyright. First year is "the year of first external delivery" and the second is "the first external distribution of the latest substantial changes made in a module in either a source or object form".

PR #564, opened 2023-03-22
test: rewrite DatabaseCompare natively

tomblench:
naive implementation, no attempt at optimisation

## Checklist

## Description

Test only changes, add javascript methods to replace `DatabaseCompare` java utility

## Approach

- add compare.js
- change tests to use native compare
- remove all references to DatabaseCompare and related env vars.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

Change affects test code so manual verification will need to be performed before merging

## Monitoring and Logging

- "No change"


package-lock.json
```diff
@@ -1,8 +1,4175 @@
 {
   "name": "@cloudant/couchbackup",
   "version": "2.9.8-SNAPSHOT",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
ricellis, line 4: We've had issues in the past with the peer dependencies working correctly in Node 14/NPM 6 with the v2 lockfiles from newer versions. I think you'll need to regenerate this with the older version.

test/compare.js
```diff
@@ -0,0 +1,96 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+const _ = require('lodash');
```
ricellis, line 17: I wonder if it might be nicer to just import the specific functions we need
```suggestion
const chunk = require("lodash/chunk");
const difference = require("lodash/difference");
const forOwn = require("lodash/forOwn");
const isEmpty = require("lodash/isEmpty");
const union = require("lodash/union");
```

package-lock.json
tomblench: thanks for spotting, done in 66a8842

ricellis (approved):

PR #628, opened 2023-10-13
Modernize citestutils and test cases

emlaver:
## Checklist

## Description

Modernize functions in `citestutils` and non-unit backup/restore test cases:
- Return promises and convert functions to async
- Remove mocha `done` callback


test/citestutils.js
```diff
@@ -134,6 +134,25 @@ function testBackup(params, databaseName, outputStream, callback) {
...
+            reject(err);
+          }
+          process.stderr.removeListener('data', listener);
+        }
+      };
+      backup.childProcess.stderr.on('data', listener);
+    });
+    promises.push(listenerPromise);
+  }
+
```
ricellis, line 155: Is this is a bad merge? It's duplicating 118-135 but outside of the CLI clause.

test/citestutils.js
```diff
@@ -170,11 +189,11 @@ function testBackup(params, databaseName, outputStream, callback) {
   }
 
   // Wait for the promises and then assert
-  return Promise.all(promises)
+  return await Promise.all(promises)
```
ricellis, line 192: I don't think this should `await` since the promise chain is getting returned to the caller (and eventually to mocha which will handle any necessary waiting)
```suggestion
  return Promise.all(promises)
```

test/citestutils.js
```diff
@@ -282,12 +296,12 @@ function testRestore(params, inputStream, databaseName, callback) {
   promises.unshift(pipeline(pipelineStreams));
 
   // Wait for the all the promises to settle and then assert based on the process promise
-  return Promise.allSettled(promises)
+  return await Promise.allSettled(promises)
```
ricellis, line 299: as for backup case there shouldn't be a need to `await` here
```suggestion
  return Promise.allSettled(promises)
```

test/citestutils.js
```diff
@@ -112,7 +112,7 @@ function testBackup(params, databaseName, outputStream, callback) {
         }
       });
       tail.on('error', function(err) {
-        callback(err);
+        return Promise.reject(err);
```
ricellis, line 115: If this error is encountered we'll get a "Unhandled promise rejection" error in Node because this `on` event listener is separate from all the promise chains.
We could probably set up a new promise that rejects for the `error` listener and resolves for the `close` event and add it to the `promises` array.
That said I don't think this is very likely anyway, we could probably just remove this `error` listener and just let Node surface an unhandled error event if it happens (which newer Node versions do anyway IIRC).

test/citestutils.js
```diff
@@ -299,131 +313,87 @@ function testRestore(params, inputStream, databaseName, callback) {
...
-function testDirectBackupAndRestore(params, srcDb, targetDb, callback) {
+async function testDirectBackupAndRestore(params, srcDb, targetDb) {
   // Allow a 64 MB highWaterMark for the passthrough during testing
   const passthrough = new PassThrough({ highWaterMark: 67108864 });
   const backup = testBackup(params, srcDb, passthrough, () => {});
   const restore = testRestore(params, passthrough, targetDb, () => {});
-  Promise.all([backup, restore]).then(() => {
-    dbCompare(srcDb, targetDb, callback);
-  }).catch((err) => callback(err));
+  return await Promise.all([backup, restore]).then(() => {
```
ricellis, line 350:
```suggestion
  return Promise.all([backup, restore]).then(() => {
```

test/citestutils.js
```diff
@@ -299,131 +313,87 @@ function testRestore(params, inputStream, databaseName, callback) {
...
     }
   };
 
   if (params.opts && params.opts.output) {
-    testBackup(params, srcDb, new PassThrough(), resume);
+    return testBackup(params, srcDb, new PassThrough()).then(() => { return resume(); });
   } else {
-    testBackupToFile(params, srcDb, backupFile, resume);
+    return testBackupToFile(params, srcDb, backupFile).then(() => { return resume(); });
   }
```
ricellis, line 388: I think a dbCompare has gone missing in here somewhere. I think though that with promises we can make this a  lot easier to read with fewer functions e.g. something like:
```js
  return Promise.resolve()
    .then(() => {
      // First backup with an abort
      if (params.opts && params.opts.output) {
        return testBackup(params, srcDb, new PassThrough());
      } else {
        return testBackupToFile(params, srcDb, backupFile);
      }
    }).then(() => {
      // Remove the abort parameter and add the resume parameter
      delete params.abort;
      params.opts.resume = true;
      // Resume the backup
      if (params.opts && params.opts.output) {
        return testBackup(params, srcDb, new PassThrough());
      } else {
        return testBackupToFile(params, srcDb, backupFile);
      }
    }).then(() => {
      // Restore the backup
      return testRestoreFromFile(params, backupFile, targetDb);
    }).then(() => {
      // Now compare the restored to the original for validation
      return dbCompare(srcDb, targetDb);
    });
```

ricellis (changes requested):
This is looking good, a few suggestions, plus some copyright updates needed.

test/ci_e2e.js
```diff
@@ -19,18 +19,19 @@ delete require.cache[require.resolve('./citestutils.js')];
 const u = require('./citestutils.js');
 
 [{ useApi: true }, { useApi: false }].forEach(function(params) {
-  describe(u.scenario('End to end backup and restore', params), function() {
-    it('should backup and restore animaldb', function(done) {
+  describe(u.scenario('end to end backup and restore', params), function() {
+    it('should backup and restore animaldb e2e', async function() {
```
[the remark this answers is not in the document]
emlaver: Yes, this was for local testing that I forgot to revert.

ricellis (approved):

PR #630, opened 2023-10-18
Modernize unit tests

emlaver:
## Checklist

## Description

Modernize unit test cases:
- Return promises and convert functions to async
- Remove mocha `done` callback

### DIFF
--- a/test/app.js
+++ b/test/app.js
@@ -1,4 +1,4 @@
-// Copyright © 2017, 2018 IBM Corp. All rights reserved.
+// Copyright © 2017, 2023 IBM Corp. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -19,6 +19,8 @@ const assert = require('assert');
 const backup = require('../app.js').backup;
 const fs = require('fs');
 const nock = require('nock');
+const util = require('util');
+const backupPromise = util.promisify(backup);
 
 const goodUrl = 'http://localhost:5984/db';
 // The real validateArgs function of app.js isn't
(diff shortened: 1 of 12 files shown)


test/writer.js
```diff
@@ -41,39 +41,51 @@ describe('#unit Check database restore writer', function() {
...
-      .pipe(writer(db, 500, 1, noopEmitter))
-      .on('error', function(err) {
-        done(err);
-      })
-      .on('finished', function(data) {
-        assert.strictEqual(data.total, 15);
-        assert.ok(nock.isDone());
-        done();
-      });
+    return new Promise((resolve, reject) => {
```
ricellis, line 49: As you mentioned to me I think we can use pipeline here, it would look something like:
```
    const w = writer(db, 500, 1, noopEmitter);
    return Promise.all([pipeline(fs.createReadStream('./test/fixtures/animaldb_expected.json'), liner(), w),
      once(w, 'finished').then((data) => {
        assert.strictEqual(data.total, 15);
        assert.ok(nock.isDone());
      })]);
```

test/writer.js
```diff
@@ -41,39 +41,51 @@ describe('#unit Check database restore writer', function() {
...
 
-    testLinestream()
-      .pipe(writer(db, 500, 1, noopEmitter))
-      .on('error', function(err) {
-        assert.strictEqual(err.name, 'Unauthorized');
-        assert.strictEqual(err.message, 'Access is denied due to invalid credentials.');
-        assert.ok(nock.isDone());
-        done();
-      });
+    return new Promise((resolve, reject) => {
```
ricellis, line 72: In the rejection case I think a pipeline like this could work:
```
    const w = writer(db, 500, 1, noopEmitter);
    return assert.rejects(
      pipeline(fs.createReadStream('./test/fixtures/animaldb_expected.json'), liner(), w),
      (err) => {
        assert.strictEqual(err.name, 'Unauthorized');
        assert.strictEqual(err.message, 'Access is denied due to invalid credentials.');
        assert.ok(nock.isDone());
        return true;
      });
```

ofc the pipeline is the `const { pipeline } = require('node:streams/promises');` form.

ricellis (commented):
For `logfilegetbatches` `logfilesummary` and `spoolchanges` it might be easier to read to `utils.promisify` the functions under test and then return it with assertions in a `then` instead of needing so many `resolve`/`reject` `try`/`catch` blocks around.
We may well tweak them again as we change the blocks underneath so if you don't want to change them I can live with it.

Other comments inline.

test/fatalerrors.js
```diff
@@ -98,12 +91,10 @@ function restoreHttpError(opts, errorName, errorCode, done) {
...
-    after('Reset process data', function(done) {
+    after('Reset process data', function() {
       process.env = processEnvCopy;
       if (!params.useApi) {
-        proxy.close(done);
-      } else {
-        done();
+        proxy.close();
       }
     });
```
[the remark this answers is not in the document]
ricellis: The "revert" of this has left an extra `proxy.close();` in place

test/app.js
```diff
@@ -19,33 +19,50 @@ const assert = require('assert');
...
+const backupPromise = util.promisify(backup);
 
 const goodUrl = 'http://localhost:5984/db';
 // The real validateArgs function of app.js isn't
 // exported - so we call the exported backup method
 // instead. We don't get as far as a real backup when
 // testing error cases. For success cases we nock the
 // goodUrl and
-const validateArgs = function(url, opts, callback) {
+const validateArgs = async function(url, opts, callback) {
```
ricellis, line 31: I don't think we should be using a callback in an async function, even if it is just a way of passing the error around (the naming `callback` is confusing if nothing else because of the history here).

I think we can simplify this by taking advantage of `assert.rejects`: a proposal in https://github.com/IBM/couchbackup/commit/5f8b4e3d8ce517024d6aa299d1b98aab35a805ca

ricellis (changes requested):
nearly there!

ricellis (approved):

PR #641, opened 2023-10-27
test: improve error handling in fatal errors suite

### LINKED ISSUE #640: CLI test failures can cascade due to proxy errors

Run the fatal errors test suite for CLI with a mismatched nock request.

The test to fail.

### 3. What actually happened

A timeout followed by another test failure.
The proxy used for CLI tests doesn't handle errors, so the nock mismatch prevents the proxy responding and causes an unhandled error. Subsequent proxied tests then fail with connection errors because the proxy is unresponsive.

## Environment details
    > main
- Node.js version
    > 18

ricellis:
## Checklist

- [x] Added tests for code changes _or_ test/build only changes - test changes
- [x] Updated the change log file (`CHANGES.md`) _or_ test/build only changes - test changes

## Description

Improve error handling in fatal errors suite.
Fixes #640 

## Approach

* Make `testPromiseWithAssertNock` and use it from all tests so the nock assertion can be controlled in one place.
* Move the `assertNock` from `then` to `finally` so that in cases where the test failed with an error we can see if nock was complete.
* Add an `error` event listener for the CLI test proxy and write a `400` result when errors are encountered. This prevents the proxy server blocking subsequent requests.
* Add a `no match` event listener for `nock` in the test `before` (and remove it in the `after`). This way any unmatched requests can be logged which makes debugging easier.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- Modified existing tests as outlined above to fix #640.

## Monitoring and Logging

- "No change"

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "package-lock.json|test/fixtures|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2023-10-20T14:17:37Z",
+  "generated_at": "2023-10-26T17:26:38Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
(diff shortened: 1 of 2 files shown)


emlaver (approved):
Looks good!

PR #657, opened 2023-11-23
test: intermittent failures in fatal error tests

ricellis:
## Checklist

- [x] Added tests for code changes _or_ test/build only changes - test only
- [x] Updated the change log file (`CHANGES.md`) _or_ test/build only changes - test only

Run the fatal error tests repeatedly on a busy system.

All tests to pass.

### 3. What actually happened

The test intermittently fail (particularly the CLI restore tests).
These tests fork a new Node process and also make real requests against mocks (via a proxy).
This amount of work seems to sometimes trigger the default 2s test timeout on busy systems.

## Approach

Inrease the test timeout for these tests to 10 s.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- Modified existing tests as outlined above. I tested 15 builds with the increased timeout and did not see any failures, previously the timeout failure would occur approximately 1 in 5.

## Monitoring and Logging

- "No change"

### DIFF
--- a/test/fatalerrors.js
+++ b/test/fatalerrors.js
@@ -76,6 +76,10 @@ async function restoreHttpError(opts, errorName, errorCode) {
 
 [{ useApi: true }, { useApi: false }].forEach(function(params) {
   describe(u.scenario('#unit Fatal errors', params), function() {
+    // These tests do real requests with mocks and if they run slowly
+    // the 2 second default mocha timeout can be insufficient, use 10 s
+    this.timeout(10000);
+
     let processEnvCopy;
     let proxy;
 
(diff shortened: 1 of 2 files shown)


test/fatalerrors.js
```diff
@@ -76,6 +76,10 @@ async function restoreHttpError(opts, errorName, errorCode) {
 
 [{ useApi: true }, { useApi: false }].forEach(function(params) {
   describe(u.scenario('#unit Fatal errors', params), function() {
+    // These tests do real requests with mocks and if they run slowly
+    // the 2 second default timeout can be insufficient, use 10 s
```
mojito317, line 80: nit
```suggestion
    // the 2 second default mocha timeout can be insufficient, use 10 s
```

mojito317 (approved):
LGTM! +1
Maybe being more specific which timeout is being set would be helpful for the future, but I'm also ok with the current version, so it's up to you.

PR #659, opened 2023-11-27
refactor: writable wrappers

ricellis:
## Checklist

- [x] Updated the change log file (`CHANGES.md`) _or_ test/build only changes - merging to `modernization` changes will be updated later.

## Description

Add `Writable` instances that delegate to another `Writable`.

Fixes i259

## Approach

* Re-ordered transforms and renamed test describe blocks
* Added 2 new classes:
    * `DelegateWritable` - a `Writable` implementation that simply delegates to another `Writable`. Special handling is added for `process.stdout`. It is not "endable" while the process is running so we need to detect that case to use it at the end of a pipeline.
    * `WritablePassThrough` - a special case of the `SideEffect` pass through that uses a `DelegateWritable` for performing side effects. This allows for the case of writing a log file from a middle stage of a pipeline.
* Updated `spoolchanges` to use the `DelegateWritable` as much of the functionality was similar to the `LogWriter` there.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- Added new tests in `test/transform.js` for:
    - `DelegateWritable`
    - `WritablePassThrough`

## Monitoring and Logging

- Added logging throughout new classes.

### DIFF
--- a/includes/spoolchanges.js
+++ b/includes/spoolchanges.js
@@ -15,10 +15,9 @@
 
 const fs = require('fs');
 const error = require('./error.js');
-const { BatchingStream, MappingStream } = require('./transforms.js');
+const { BatchingStream, DelegateWritable, MappingStream } = require('./transforms.js');
 const debug = require('debug')('couchbackup:spoolchanges');
 const { ChangesFollower } = require('@ibm-cloud/cloudant');
-const { Writable } = require('node:stream');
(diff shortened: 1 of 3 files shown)


eiri (approved):

emlaver (approved):
Looks good!

PR #664, opened 2023-11-30
refactor: Promisify backup and restore pre-flight checks

eiri:
This is an async rewrite of pre-flight checks for backup and restore functions to do a pre-work to conversion from internal event emitter + callbacks to streams rewrite.

The steps done: 

- Simplify and DRY `validateArgs` in `app.js`
- Convert all validation functions to `async`, convert backup/restore to use them in promise chain fashion.

### DIFF
--- a/app.js
+++ b/app.js
@@ -56,28 +56,27 @@ function isSafePositiveInteger(x) {
  * @param {boolean} isIAM - A flag if IAM authentication been used.
  * @returns Boolean true if all checks are passing.
  */
-function validateURL(url, isIAM) {
+async function validateURL(url, isIAM) {
   if (typeof url !== 'string') {
     throw new OptionError('Invalid URL, must be type string');
   }
   // Validate URL and ensure no auth if using key
-  let urlObject
   try {
-    urlObject = new URL(url);
+    const urlObject = new URL(url);
+    // We require a protocol, host and path (for db), fail if any is missing.
+    if (urlObject.protocol !== 'https:' && urlObject.protocol !== 'http:') {
+      throw new OptionError('Invalid URL protocol.');
+    }
+    if (!urlObject.pathname || urlObject.pathname === '/') {
+      throw new OptionError('Invalid URL, missing path element (no database).');
+    }
+    if (isIAM && (urlObject.username || urlObject.password)) {
+      throw new OptionError('URL user information must not be supplied when using IAM API key.');
+    }
   } catch (err) {
     throw error.wrapPossibleInvalidUrlError(err);
   }
-  // We require a protocol, host and path (for db), fail if any is missing.
-  if (urlObject.protocol !== 'https:' && urlObject.protocol !== 'http:') {
-    throw new OptionError('Invalid URL protocol.');
-  }
-  if (!urlObject.pathname || urlObject.pathname === '/') {
-    throw new OptionError('Invalid URL, missing path element (no database).');
-  }
-  if (isIAM && (urlObject.username || urlObject.password)) {
-    throw new OptionError('URL user information must not be supplied when using IAM API key.');
-  }
-  return true
+  return true;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


app.js
```diff
@@ -49,97 +50,144 @@ function isSafePositiveInteger(x) {
...
+      case 'number':
+        if (typeof val !== 'undefined' && !isSafePositiveInteger(val)) {
+          throw new OptionError(`Invalid ${rule.key} option, must be a positive integer in the range (0, MAX_SAFE_INTEGER]`);
+        }
+        break;
+      case 'enum':
+        if (typeof val !== 'undefined' && rule.values.indexOf(val) === -1) {
+          // this is a shortcut. it should be join on rule.values,
+          // but we have just one enum, so hardcode will do for now
+          throw new OptionError('Invalid mode option, must be either "full" or "shallow"');
```
ricellis, line 121: Agreed, not really necessary when we have only one, but I think we can do this fairly easily:
```suggestion
          throw new OptionError(`Invalid mode option, must be one of: ${JSON.stringify(rule.values).slice(1,-1)`);
```

app.js
```diff
@@ -225,109 +284,101 @@ module.exports = {
...
+            targetStream.write('', 'utf8', emitFinished);
+          } else {
+            // If we're writing to a file, end the writes and register the
+            // emitFinished function for a callback when the file stream's finish
+            // event is emitted.
+            targetStream.end('', 'utf8', emitFinished);
+          }
+        });
+      })
+      .catch(callback);
```
ricellis, line 380: Do you think this should move up to after the `.then((backupDB) => validateBackupDb(backupDB))`. My reasoning being that errors that happen in the final `then` block where the backup is actually done are being passed directrly to the callbacks.
I don't think there is, but I can't really figure out if there is a circumstance where we'd end up catching an error that we should otherwise let through. Moving it up would just eliminate that risk. I think once we have the pipeline promise here it won't matter anyway because then we definitely will want a catch at the end, so maybe we can just leave it temporarily. WDYT?

app.js
```diff
@@ -344,64 +395,57 @@ module.exports = {
...
+                callback(null, obj);
+              });
+            }
           }
-        }
-      );
-    });
+        );
+      })
+      .catch(callback);
```
ricellis, line 447: as above

app.js
eiri: This is missing "or" though, changing it for no-space comma separated. Something like `rule.values.map(x => JSON.stringify(x)).join(' or ')` is closer to what we have here, but that breaks on (future) 3 elements enum, so I just opted "we'll think about if we have to"

app.js
```diff
@@ -49,97 +50,144 @@ function isSafePositiveInteger(x) {
...
+  }
+  const rules = [
+    { key: 'iamApiKey', type: 'string' },
+    { key: 'log', type: 'string' },
+    { key: 'output', type: 'string' },
+    { key: 'bufferSize', type: 'number' },
+    { key: 'parallelism', type: 'number' },
+    { key: 'requestTimeout', type: 'number' },
+    { key: 'mode', type: 'enum', values: ['full', 'shallow'] },
+    { key: 'resume', type: 'boolean' }
```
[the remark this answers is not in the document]
eiri: Well, i did the change mostly for consistency, we throw "Invalid iamApiKey option, must be type string" for string, but doing humanized version for numbers. I can do a special case here and match what we have before, that's not that hard.

app.js
eiri: I'm ok with moving it up, after all `catch` is just a syntax sugar version of `then(null, err)`. I put it at the bottom with the idea that we are going to fix internalBackup / internalRestore next, so this will be single place we are using  `callback` and since we are doing feature branch it's safe to have the final form here.

Let's move it for now, put it down later.

app.js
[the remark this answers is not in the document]
eiri: Done in https://github.com/IBM/couchbackup/pull/664/commits/d95662c3eed662c74d7cc2add47beb682fc1ef1d

app.js
eiri: Not _that_ simple, still doable. Done in https://github.com/IBM/couchbackup/pull/664/commits/d95662c3eed662c74d7cc2add47beb682fc1ef1d

app.js
eiri: Done in https://github.com/IBM/couchbackup/pull/664/commits/5d685ffb5a403c97b91d557d9d1ca1bb3f8013b9

app.js
eiri: Well, that was educational. It's obvious in hindsight, but `catch` doesn't break a promise chain, so just moving it wasn't enough, it should explicitly throw (or `Promise.reject`) after calling `callback`.

app.js
ricellis: >This is missing "or" though

that's why I changed it to "one of", but anyway this is fine.

app.js
ricellis: Sorry, yeah, that should have been obvious. Mixing promises and callbacks, just don't do it 😄 
Anyway, as we said, this is temporary and won't be shipped and it'll be cleaner anyway once we join all the things together.

ricellis (approved):

emlaver (approved):

PR #669, opened 2023-12-11
refactor: backup pipeline

ricellis:
## Checklist

## Description

Migrate the full backup functionality to a streams pipeline.

Fixes i264 and i265.

## Approach

* refactor: resume log file handling
    * Replace JS object with `Map` for storing state
    * `logFileSummary` replaced with a pipeline that reads log file metadata only and resolves with a summary of pending batch numbers
    * `logFileGetBatches` returns an array of streams that can be composed into a pipeline that maps log file data to pending backup batches
* refactor: spoolchanges for backup pipeline & fix: spool changes uses non-existent error name
    * changed to return an array of streams that can be composed into a pipeline that streams changes and adds pending batches to a log file
    * fix a bug where an invalid change used an error name not present in `errors.js`
    * make `LogWriter` a `WritableWithPassThrough` so that it can log a pending batch and pass on the backupBatch to be used later in the pipeline
* refactor: full backup as pipeline
    * Full backup in app.js becomes a promise that when resolved/rejected triggers the user callback (existing code moves into shallow backup only for now)
    * replace the backupFull function with a pipeline promise composed of a source from either a log file resume or spool changes followed by complete backup batch fetch/write/log.
* refactor: event emitters
     * add a `SideEffect` to emit `changes` events after they are received and logged
     * add a `postWriteFn` to the backup destination `Writable` so that it emits the `written` event after the batch has been written to the file
* fix: restore totals
    * fixes a bug where restored totals were out of order

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- Modified existing tests:
    - log file tests updated to use new interfaces described above
    - spool changes tests
        - updated to use interface above with a common function to compose a pipeline for the tests and convert the errors caught in the pipeline
        - updated fatalerrors test to use an invalid change
    - test: fix write error test
        - this test was not working as designed, the `EACCES` error was being thrown from the `createWriteStream` and not from inside backup code. This also meant the backup was running in the background and could cause subsequent tests to fail. Just pass a `Readable` stream into the API instead to prove the backup fails if it can't write.

## Monitoring and Logging

- Some debug logging is replaced by new logging in transforms and backupMappings.

### DIFF
--- a/app.js
+++ b/app.js
@@ -423,7 +423,8 @@ module.exports = {
           objectMode: true,
           write: (restoreBatch, encoding, cb) => {
             debug(' restored ', restoreBatch.total);
-            total = restoreBatch.total;
+            // order not guaranteed
+            if (total < restoreBatch.total) total = restoreBatch.total;
             try {
               ee.emit('restored', restoreBatch);
             } finally {
--- a/app.js
+++ b/app.js
@@ -20,11 +20,11 @@
  */
 
 const backupFull = require('./includes/backup.js');
+const backupShallow = require('./includes/shallowbackup.js');
 const defaults = require('./includes/config.js').apiDefaults;
 const error = require('./includes/error.js');
 const request = require('./includes/request.js');
 const restoreInternal = require('./includes/restore.js');
-const backupShallow = require('./includes/shallowbackup.js');
 const debug = require('debug')('couchbackup:app');
 const events = require('events');
 const fs = require('fs');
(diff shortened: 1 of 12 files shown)


test/spoolchanges.js
```diff
@@ -32,6 +36,18 @@ const db = request.client(`${url}/${dbName}`, { parallelism: 1 });
...
 
+function changes(bufferSize, tolerance) {
+  // Make a pipeline of the spool changes source streams
+  return pipeline(...spoolchanges(db, '/dev/null', bufferSize, tolerance),
+  // add a mapping to string and send to /dev/null as we don't care about the output
+    new MappingStream(JSON.stringify), fs.createWriteStream('/dev/null'))
+    // Historically spool changes itself could return an error, but
+    // now it returns streams to be made a pipeline elsewhere.
+    // As such a conversion takes place in the pipeline level catch, we reproduce
+    // that here by calling the converResponseError function.
```
emlaver, line 47:
```suggestion
    // that here by calling the convertResponseError function.
```

emlaver (commented):
Looks great

includes/spoolchanges.js
```diff
@@ -29,29 +28,24 @@ const { pipeline } = require('node:stream/promises');
  * @param {number} bufferSize - the number of changes per batch/log line
```
emlaver, line 28: On line 26, type should be changed to `{object}`.

emlaver:
Copyright update for `includes/logfilesummary`

app.js
```diff
@@ -304,88 +304,85 @@ module.exports = {
...
-        // an interruption of the previous backup. If the backup was clean this
-        // will cause an empty line that will be gracefully handled by the restore.
-        if (opts.resume) {
-          targetStream.write('\n');
-        }
+          // if there is an error writing to the stream, call the completion
+          // callback with the error set
+          const listenerErrorIndicator = { errored: false };
+          addEventListener(listenerErrorIndicator, targetStream, 'error', function(err) {
+            debug('Error ' + JSON.stringify(err));
```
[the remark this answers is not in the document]
ricellis: These two `Error` debug statements are just moved shallow code that will be repaced by the shallow pipeline so I don't think they really need updating. Were there other "net new" debug statements you weren't happy with?

includes/backup.js
```diff
@@ -13,267 +13,101 @@
...
- * @param {any} parallelism - number of concurrent requests to make
- * @param {any} log - log file to drive downloads from
- * @param {any} batches - batches to download
- * @param {any} ee - event emitter for progress. This funciton emits
- *  received and error events.
- * @param {any} start - time backup started, to report deltas
- * @param {any} grandtotal - count of documents downloaded prior to this set
- *  of batches
- * @param {any} callback - completion callback, (err, {total: number}).
+ * @param {string} db - `@cloudant/cloudant` DB object for source database.
```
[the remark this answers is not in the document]
ricellis: added to i313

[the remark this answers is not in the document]
ricellis: doc updated in 32cbd5dcce26200a5e4e33064fc21f5a76615bdb

includes/backup.js
```diff
@@ -13,267 +13,101 @@
...
+ * @param {number} options - backup configuration
+ * @param {Writable} targetStream - destination for the backup contents
+ * @param {EventEmitter} ee - user facing event emitter
+ * @returns pipeline promise that resolves for a successful backup or rjects on failure
```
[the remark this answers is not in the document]
ricellis: fixed in 32cbd5dcce26200a5e4e33064fc21f5a76615bdb

includes/backup.js
```diff
@@ -13,267 +13,101 @@
...
-
-  return ee;
-};
+const { MappingStream, WritableWithPassThrough, DelegateWritable, SideEffect } = require('./transforms.js');
 
 /**
  * Validate /_bulk_get support for a specified database.
  *
  * @param {string} db - nodejs-cloudant db
  * @param {function} callback - called on completion with signature (err)
```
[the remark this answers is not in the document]
ricellis: fixed in 32cbd5dcce26200a5e4e33064fc21f5a76615bdb

test/spoolchanges.js
ricellis: Fixed in 966cdb5b159ed1866ae7ea7abc6ea2fdb25612e7

includes/spoolchanges.js
ricellis: 966cdb5b159ed1866ae7ea7abc6ea2fdb25612e7

app.js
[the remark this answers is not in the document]
emlaver: Ok good point, let's leave it as-is.  I don't have issues with any of the other debug statements.

emlaver (approved):

app.js
```diff
@@ -304,88 +304,85 @@ module.exports = {
...
+          // an interruption of the previous backup. If the backup was clean this
+          // will cause an empty line that will be gracefully handled by the restore.
+          if (opts.resume) {
+            targetStream.write('\n');
           }
-        });
-      });
+
+          return backupFull(backupDB, opts, targetStream, ee)
+            .then(callback); // TODO move after shallow backup is refactored
```
eiri, line 382: This bit confuses me - what does it do?

includes/backup.js
```diff
@@ -13,267 +13,100 @@
...
  */
-function validateBulkGetSupport(db, callback) {
-  db.service.postBulkGet({ db: db.db, docs: [] }).then(() => { callback(); }).catch(err => {
-    err = error.convertResponseError(err, function(err) {
-      switch (err.status) {
-        case undefined:
+async function validateBulkGetSupport(db) {
+  return db.service.postBulkGet({ db: db.db, docs: [] })
+    .catch(err => {
+      err = convertResponseError(err, function(err) {
```
eiri, line 33: So we throw from inside error factory and doing not working reassign on `err` which we doesn't return anyway here? This looks wrong, it seems that instead the error factory should return composed error and catch should return or re-throw it as `throw convertResponseError` instead.  Honestly this would be less confusing as `try {} catch` around `postBulkGet` instead.

includes/backup.js
```diff
@@ -13,267 +13,100 @@
...
+            total += backupBatch.docs.length;
+            ee.emit('written', { total, time: (new Date().getTime() - start) / 1000, batch: backupBatch.batch });
+          } // post write function emits the written event
+        ) // DelegateWritable writes the log file done lines
+      );
+    })
+    .then(() => {
+      ee.emit('finished', { total });
+    })
+    .catch((e) => { throw convertResponseError(e); });
```
eiri, line 111: That's a nit, but we don't need curly on one-liners or parens on one-args, so this can be written as `.catch(e => throw convertResponseError(e))`

Same with above, btw, are curlies there just to suppress return from `ee.emit` ?

includes/spoolchanges.js
```diff
@@ -62,27 +56,31 @@ module.exports = async function(db, log, bufferSize, tolerance = 600000) {
...
   };
 
+  const mapChangesBatchToBackupBatch = function(changesBatch) {
+    return { command: 't', batch: batch++, docs: mapChangesToIds(changesBatch) };
+  };
+
+  const mapBackupBatchToPendingLogLine = function(backupBatch) {
+    totalBuffer += backupBatch.docs.length;
+    debug('writing', backupBatch.docs.length, 'changes to the backup log file with total of', totalBuffer);
+    return `:t batch${backupBatch.batch} ${JSON.stringify(backupBatch.docs)}\n`;
```
eiri, line 71: What this `:t` does?

app.js
ricellis: This calls the user supplied `callback` (which is a user supplied function when using the API or a function from the bin scripts for CLI usage). We have to have a callback to retain the existinng API contract.

Right now there is a split between shallow and full backup in `app.js`. In the shallow half is still doing stuff via an internal event emitter, including the user API callback.
In the full version we need to callback after the pipeline promise completes. That's what this is doing.

Once we do shallow pipeline as well then this will be able to move to the end of the promise chain because it will all be able to happen in a single place.

includes/spoolchanges.js
ricellis: `:t` is the log file prefix for "todo" i.e. pending batches

app.js
eiri: But our `backupFull` doesn't return anything on success, it just emits `finished` and voids and we have `catch` to  call `callback` on failure, so this is no-op? Sorry, I should've elaborate on that this is the part that confuses me

eiri: I think I finally got it, this is calling `callback(null, undefined)` on successful finish, so user can run theirs' side-effects? weird, but ok

includes/backup.js
ricellis: For the former: 93b4a2dd0acf7dc4d0e80cfb2a25eefdd01bde5b
For the latter:
`.catch(e => throw convertResponseError(e));`
gives me an
"Unexpected token throw"

eiri: That's weird, that really should be a syntax only thing. Is it because event emitter now returns and that interpreted as promise reject and passed as `e`? I see `the boolean true was thrown, throw an Error :)` in CI and `emit` returns `true` when there are listeners

ricellis: ok dropping it, there is some follow on cleanup coming anyway

PR #678, opened 2023-12-19
fix: use bufferSize for seqInterval

ricellis:
## Checklist

- [x] Updated the change log file (`CHANGES.md`) _or_ test/build only changes - merging to `modernization`

## Description

fix: use bufferSize for seqInterval

Fixes part of i312

## Approach

Adjust the `seqInterval` to match the `bufferSize` - this should mean we get at least one seq ID per batch we backup (unless the database is very small i.e. smaller than one batch).

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- No new tests because code path is covered by existing tests, changing the seqInterval does not impact any other behaviors at present.

## Monitoring and Logging

- `:changes_complete` log entries are more likely to have a sequence ID.

### DIFF
--- a/includes/spoolchanges.js
+++ b/includes/spoolchanges.js
@@ -28,7 +28,7 @@ const { ChangesFollower } = require('@ibm-cloud/cloudant');
  * @param {number} bufferSize - the number of changes per batch/log line
  * @param {number} tolerance - changes follower error tolerance
  */
-module.exports = function(dbClient, log, bufferSize, tolerance = 600000) {
+module.exports = function(dbClient, log, bufferSize = 500, tolerance = 600000) {
   let lastSeq;
   let batch = 0;
   let totalBuffer = 0;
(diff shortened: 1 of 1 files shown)


includes/spoolchanges.js
```diff
@@ -73,7 +73,7 @@ module.exports = function(dbClient, log, bufferSize, tolerance = 600000) {
 
   const changesParams = {
     db: dbClient.dbName,
-    seqInterval: 10000
+    seqInterval: bufferSize
```
eiri, line 76: This is now coming from outside, so potentially can be undefined. Do we maybe want to default it to an old value, similar how we do with `tolerance`?

ricellis: Good point. Since we apply a default value for `bufferSize` it will never be `undefined` from usages in the backup code. However, it is possible that test code could call it `undefined`, so I've added the same as our normal default `500` just to be safe.

emlaver (approved):

eiri (approved):

PR #682, opened 2023-12-19
fix: additional resume options validation

ricellis:
## Checklist

- [x] Updated the change log file (`CHANGES.md`) _or_ test/build only changes - merging to `modernization`

## Description

Add some additional options validation for the `resume` cases.

Fixes part of i312.

## Approach

* Add new `LogFileExists` error (code 23).
* Rework `validateLogOnResume` function to split into 3 parts (no options, resuming and not resuming) with additional checks in the latter 2 parts relating to the log file.
* Add a bufferSize warning line to `validateLogOnResume`.

## Schema & API Changes

- Backup error thrown if a specified log file exists and it is a new backup.

## Security and Privacy

- "No change"

## Testing

- Added new tests:
    - assert the buffer size warning is printed when using `resume` and `bufferSize` options.
    - assert the `LogFileExists` error is thrown when not using `resume` and specifying an existing log file.
- Modifed existing tests to avoid creating a log file in advance of the backup because that is now an error condition. Instead we have a watcher on the folder and then start the log file tail after the log file appears.

## Monitoring and Logging

-  Outputs a warning if resuming and specifying `bufferSize`.

### DIFF
--- a/app.js
+++ b/app.js
@@ -173,6 +173,10 @@ async function validateLogOnResume(opts) {
     } else if (!logFileExists) {
       throw new BackupError('LogDoesNotExist', 'To resume a backup, the log file must exist');
     }
+    if (opts.bufferSize) {
+      // Warn that the bufferSize is already fixed
+      console.warn('WARNING: the original backup "bufferSize" applies when resuming a backup.');
+    }
   } else {
     // Not resuming
     if (logFileExists) {
--- a/app.js
+++ b/app.js
@@ -159,15 +159,26 @@ async function shallowModeWarnings(opts) {
  */
 
 async function validateLogOnResume(opts) {
-  if (!opts || !opts.resume) {
+  const logFileExists = opts && opts.log && fs.existsSync(opts.log);
+  if (!opts || opts.mode === 'shallow') {
+    // No opts specified, defaults will be populated.
+    // In shallow mode log/resume are irrelevant and we'll have warned already.
     return true;
-  }
-  if (!opts.log) {
-    // This is the second place we check for the presence of the log option in conjunction with resume
-    // It has to be here for the API case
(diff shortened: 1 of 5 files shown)


emlaver (approved):

app.js
```diff
@@ -160,15 +160,30 @@ async function shallowModeWarnings(opts) {
...
+    if (!opts.log) {
+      // This is the second place we check for the presence of the log option in conjunction with resume
+      // It has to be here for the API case
+      throw new BackupError('NoLogFileName', 'To resume a backup, a log file must be specified');
+    } else if (!logFileExists) {
+      throw new BackupError('LogDoesNotExist', 'To resume a backup, the log file must exist');
+    }
+    if (opts.bufferSize) {
+      // Warn that the bufferSize is already fixed
+      console.warn('WARNING: the original backup "bufferSize" applies when resmuming a backup.');
```
eiri, line 179:
```suggestion
      console.warn('WARNING: the original backup "bufferSize" applies when resuming a backup.');
```

app.js
```diff
@@ -178,16 +193,22 @@ async function validateLogOnResume(opts) {
  *
  * @param {string} url - URL of database.
  * @param {object} opts - Options.
+ * @param {boolean} backup - true for backup, false for restore
  * @returns Boolean true if all checks are passing.
  */
-async function validateArgs(url, opts) {
+async function validateArgs(url, opts, backup = true) {
```
eiri, line 199: Maybe `isBackup`? We use var name `backup` a lot, so it became is a bit vague

eiri (approved):
a typo and a nit, but overall looks good

app.js
ricellis: I'd copied and pasted this error straight into the test too!
Thanks for spotting it.

app.js
ricellis: good idea - I've changed it

PR #688, opened 2024-01-11
ci: increase CPU request to 1 and disable concurrent builds

emlaver:
## Checklist

## Description

The fatal error unit tests failed right before the merge of https://github.com/IBM/couchbackup/pull/662.
This most likely happened when our kube CI resources were strained and being throttled .  Increasing the CPU request to 1 and disabling concurrent builds should hopefully allow for enough resources so that the test suite doesn't time out.

fixes i316

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "package-lock.json|test/fixtures|^.secrets.baseline$",
     "lines": null
   },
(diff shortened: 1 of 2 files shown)


ricellis (approved):
+1
If the problem persists we could remove Node 21 from the matrix to reduce the resources needed. We only support the LTS versions of Node anyway.

eiri (approved):

PR #691, opened 2024-01-16
fix: handle back-pressure in Liner

ricellis:
## Checklist

## Description

Improve the `Liner` implementation in `modernization` branch to respect back-pressure. 

Part of i330.

## Approach

Change from a `Transform` implementation to a `Duplex`. Transforms are well suited to 1:1 changes, but the nature of the Liner is different. It may take many chunks being written to get an entire line (many:1) or it may produce many lines from a large chunk of a file (1:many). Applying back pressure via the `_transform` `callback` is challenging when accounting for those different situations.

Instead use a `PassThrough` as the writable side of the `Duplex` and wrap that stream in the `readline` interface. If that streams buffers are full it passes back-pressure upstream to the file via the `write`/`callback` delegation.
The readable side of the `Duplex` is an array of lines that is populated by the `line` event of the `readline` interface. When `push` is called to pass lines downstream the back-pressure is honoured by calling `pause` on the `readline` interface and setting a flag. When downstream is ready to consume again it calls `read` and our `_read` implementation reads from the buffer.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- Funcitonality is covered by existing tests. (This fix is for regressions observed in the large test suite).

## Monitoring and Logging

- Added new log group `couchbackup:liner` to facilitate debugging of this class.

### DIFF
--- a/includes/liner.js
+++ b/includes/liner.js
@@ -1,4 +1,4 @@
-// Copyright © 2017, 2023 IBM Corp. All rights reserved.
+// Copyright © 2017, 2024 IBM Corp. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


emlaver (approved):
Looks great!

eiri (approved):
lgtm

PR #693, opened 2024-01-16
fix: batching transform back-pressure

ricellis:
## Checklist

- [x] Updated the change log file (`CHANGES.md`) _or_ test/build only changes - merging to `modernization`, changes will be udpated later.

## Description

Fix batching back-pressure and reduce double-buffering on mapping streams.
Part of i330.

## Approach

* `fix: batching transform back-pressure`
    * Check the return value of the `Transform` `push` call to see if there is downstream pressure.
    * If there is, delay the `callback` from the `_transform` to pass that pressure upstream.
    * Override the `_read` method in the batching `Transform` implementation to allow calling the delayed callback when the downstream calls `_read` again (i.e. it is ready to receive more data).
    * Minimize the buffering in the batching transform by reducing the highWaterMarks
* `fix: minimize buffering on map passthroughs`
    * The `PassThrough` used in the mapping transforms would default to a buffer size of 16 objects, this could be very large batches. In practice there is no need for this stream to buffer at all instead it can directly pass from the writable side of the duplex to the buffer in the mapping operation. This is achieved by setting the highWaterMarks to `0`. They are set separately to allow for over-riding one or the other by passing `opts` (if a single `highWaterMark` is set it applies to both readable and writable regardless of their indvidual settings).

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- No new tests because functionality is tested by existing tests. (This is in response to regressions in the large test suite).

## Monitoring and Logging

- Added some additional logging in the batching transform.

### DIFF
--- a/includes/transforms.js
+++ b/includes/transforms.js
@@ -1,4 +1,4 @@
-// Copyright © 2023 IBM Corp. All rights reserved.
+// Copyright © 2023, 2024 IBM Corp. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


ricellis:
I'm going to reopen this in a different form.

PR #696, opened 2024-01-19
fix: high water marks in mapping streams

ricellis:
## Checklist

- [x] Updated the change log file (`CHANGES.md`) _or_ test/build only changes - merging to `modernization` changes will be updated later.

## Description

Apply high water marks to the mapping streams.

Fixes part of i330

## Approach

Override the default temporarily when we create mapping streams.
`Duplex.from` does not honour `highWaterMark` configuration. Instead get the current default high water mark, set a new high water mark appropriate for our use case of minimizing the memory footprint, then reset the original high water mark.

Also change the mapping streams to extend a new `DuplexMapper` so we don't have to use the same workaround multiple times.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- No new tests because the pipeline is covered by existing tests.

## Monitoring and Logging

- "No change"

### DIFF
--- a/includes/transforms.js
+++ b/includes/transforms.js
@@ -13,7 +13,7 @@
 // limitations under the License.
 'use strict';
 
-const { Duplex, PassThrough, Writable } = require('node:stream');
+const { Duplex, PassThrough, Writable, getDefaultHighWaterMark, setDefaultHighWaterMark } = require('node:stream');
 const debug = require('debug');
 
 /**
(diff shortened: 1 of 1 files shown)


eiri (approved):
lgtm

emlaver (approved):

mojito317 (approved):

Issue #113, opened 2017-06-09
Don't wait for the changes spooler to finish before starting backup doc fetches


smithsz:
Currently the backup /_bulk_get POST requests don't start until the changes spooler has finished (see https://github.com/cloudant/couchbackup/blob/master/includes/backup.js#L52).

We should start fetching document batches as soon as the spooler sends the first change.

Related, #100 - _Retry backup spool changes request on failure_.

ricellis:
We tried this as part of the modernization effort, but the downsides actually outweight the benefits (it isn't any faster because the network transfer is the bottleneck anyway). See #698 for details.

PR #698, opened 2024-01-22
fix: revert to complete spooling before backup

ricellis:
## Checklist

## Description

Use separate promises for the changes spooling and backup fetching.

Fixes i332

## Approach

We had attempted to solve #113 by putting the spooling of changes at the start of the pipeline and running straight through to backing up. The problem with this approach is that downstream back-pressure holds up the fetching of changes. This spreads the window over which couchbackup gets the list of IDs to backup over a wider window than previously. The additional time it takes to reach the end of the changes feed makes it possible for changes to be written faster than they can be backed up and potentially create a never-ending backup.

This PR reverts to the existing approach of completing the spooling of changes and then commencing the document fetches via `_bulk_get`.

* Modify `spoolchanges` to use a `DelegateWriter` for the log file as a pipeline destination instead of an intermediate stage of `WritableWithPassThrough`.
* Add a `postWrite` function to the `DelegateWriter` that can be passed in via spoolchanges init to accommodate the `batch` event emission.
* Modify the backup pipeline
    * Remove the `SideEffect` from the changes "sources" instead passsing the event emitter as outlined above
    * In the full backup logic if we are not resuming then start spooling changes and when it is complete enter the same code as resumed backups that validates changes were complete and creates a stream of batches to backup.

## Schema & API Changes

- Behavior change for backup, but this reverts to current behaviour in releases (the change only existied in the `modernization` branch - so this does not introduce any user-facing change.

## Security and Privacy

- "No change"

## Testing

- Modified existing tests for `spoolchanges` to use new init signature with a no-op postWrite function. The overall code function is already covered by existing tests.

## Monitoring and Logging

- "No change"

### DIFF
--- a/includes/backup.js
+++ b/includes/backup.js
@@ -1,4 +1,4 @@
-// Copyright © 2017, 2023 IBM Corp. All rights reserved.
+// Copyright © 2017, 2024 IBM Corp. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -20,7 +20,7 @@ const { BackupError } = require('./error.js');
 const logFileSummary = require('./logfilesummary.js');
 const logFileGetBatches = require('./logfilegetbatches.js');
 const spoolchanges = require('./spoolchanges.js');
-const { MappingStream, WritableWithPassThrough, DelegateWritable, SideEffect } = require('./transforms.js');
+const { MappingStream, WritableWithPassThrough, DelegateWritable } = require('./transforms.js');
 const allDocsGenerator = require('./allDocsGenerator.js');
 
 /**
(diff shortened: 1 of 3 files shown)


includes/backup.js
```diff
@@ -63,27 +63,29 @@ module.exports = function(dbClient, options, targetStream, ee) {
         return [
           allDocsGenerator(dbClient, options)
         ];
-      } else if (options.resume) {
-      // Resuming a backup, get the log file summary
-      // (changes complete and remaining batch numbers to backup)
+      } else {
+        // Full backup, we'll return a sream over a completed changes log file
```
emlaver, line 67:
```suggestion
        // Full backup, we'll return a stream over a completed changes log file
```

emlaver (approved):

includes/backup.js
ricellis: oh not again, I've really got to put some new shims in my keyboard 😄

PR #702, opened 2024-01-25
feat: improve handling of broken backup line during restore

emlaver:
## Checklist

## Description
In the `modernization` branch, if a restore operation encounters a broken line in the backup file it would assume that the broken line was caused by a resume and carry on doing the restore.

This work improves how broken lines are handled during backup with resume option and restore:
- During backup with resume enabled, append a `RESUME` comment marker to the last line in the backup file.
- If during the restore operation the line in the backup file fails during JSON parsing, catch the error and check if the resume marker exists.  If the marker exists then retry JSON parsing without comment, else throw an error.

tl;dr broken lines that exist in a backup file that was resumed will continue to be safely ignored during restore.  Backups with broken lines that were not resumed will now throw an exception during restore.

fixes i314

## Testing
New tests:
- three unit tests to verify resume comment marker during restore
- test to assert that an error is thrown when trying to restore file with corrupted line

Updated tests:
- `should error for a corrupted line` unit test
- `should restore corrupted animaldb to a database correctly` update fixture for test


includes/restoreMappings.js
```diff
@@ -17,6 +17,7 @@ const { BackupError } = require('./error.js');
 const debug = require('debug');
 
 const mappingDebug = debug('couchbackup:mappings');
+const RESUME_COMMENT = '##RESUME##';
```
ricellis, line 20: I've been wondering about whether we can make this comment in some kind of standards compliant way so that adding this robustness to the backup file doesn't necessarily make it incompatible with other parsers.
For example [JSON text sequences - RFC7464](https://datatracker.ietf.org/doc/html/rfc7464) provides a way to detect truncation using the `\u001E` record separator and `\n`.

We could "formalize" json-seq compatibility in the backup file by adding `\u001E` at the start of backup lines with this change. As whitespace I think it wouldn't break existing parsing (broken lines are already broken). This would also give us a way to detect that the backup line came from a version before this change, such that we could handle backwards compatibility.

This would also eliminate the (admittedly extremely low) chance of bumping into broken JSON that happens to match the comment because the `\u001E` would need to be escaped within any JSON content.

```suggestion
const RESUME_COMMENT = `\u001E${JSON.stringify('##@cloudant/couchbackup:RESUME##')}`;
```

Of course test fixtures will need updating if we go this route.

includes/restoreMappings.js
```diff
@@ -40,10 +41,21 @@ class Restore {
...
+        if (backupLine.line.slice(-RESUME_COMMENT.length) === RESUME_COMMENT) {
+          // Try to parse again with the resume marker removed
+          try {
+            arr = JSON.parse(backupLine.line.slice(0, -RESUME_COMMENT.length));
+          } catch (err) {
+            mappingDebug(`${parseError}`);
+            // Return an empty array for an ignorable line
+            return [];
+          }
+        } else {
```
ricellis, line 55:
```suggestion
        } else if (backupLine.line.charAt(0) !== '\u001E' {
          // The backup file comes from an older version of couchbackup that predates RESUME_COMMENT
          // and json-seq output. For compatibility ignore the broken JSON line assuming it was part of a
          // resume.
          console.error(`Ignoring invalid JSON on line ${backupLine.lineNumber} of backup file as it was written by an older version of couchbackup and could be a valid resume point.`);
          return [];
        } else {
```

emlaver: What about adding a version number range in the error message? It would require opening a new ticket and updating before release once we know the version number
e.g. `Ignoring invalid JSON ... written by couchbackup version < 3.0.0 and could...`

ricellis: If we have more releases before `modernization` merges they'll be patch releases on `2.9` so I think we already know it will be `2.10.0`. I've no objection to adding `< 2.10.0` to the message.

TBH I wasn't 100% sure we should output a message this obviously rather than say a debug line given we are handling it automatically.

emlaver: Fair point, we are changing the behavior to some degree from the current emit an error to print an error to console.  I'm fine with updating the message and update `console.error` -> `mappingDebug`

includes/restoreMappings.js
```diff
@@ -17,7 +17,8 @@ const { BackupError } = require('./error.js');
 const debug = require('debug');
 
 const mappingDebug = debug('couchbackup:mappings');
-const RESUME_COMMENT = ' \t ' + ' ' + '   ' + '  \t' + '\t\t' + ' '; // Special sequence of whitespace to mark a resume
+const marker = '@cloudant/couchbackup:resume';
+const RESUME_COMMENT = `${JSON.stringify({ marker })}`; // Special marker for resumes
```
eiri, line 21: Should say, I'm glad for this change. The spaces thing was cute and smart, but I had a feeling it's a bit _too_ smart and subtle.

app.js
```diff
@@ -310,16 +312,31 @@ module.exports = {
...
+          metadataToWrite = `${RESUME_COMMENT}\n`;
+        } else {
+          // Write a file header including the name, version and mode
+          debug('Will write backup file header.');
+          metadataToWrite = `${JSON.stringify({ name: pkg.name, version: pkg.version, mode: opts.mode })}\n`;
         }
-        return backup(backupDbClient, opts, targetStream, ee);
+        return new Promise((resolve, reject) => {
+          targetStream.write(metadataToWrite, 'utf-8', (err) => {
+            if (err) { reject(err); } else { resolve(backupDbClient); }
```
eiri, line 333: Oh, this use of resolve is neat.

eiri (approved):
lgtm

includes/restoreMappings.js
ricellis: Yeah after some more discussion we agreed it wasn't gaining us anything, if the backup file line was valid then the resume marker will always be on its own line anyway. If it's not on its own line then the write wasn't complete (even if the JSON was complete the missing `\n` means we didn't write the corresponding log entry) so the marker actually preventing the line from parsing is ultimately a good thing.

app.js
ricellis: I think I've played with too many PassThrough streams now, so pass through promises!

PR #727, opened 2024-03-06
ci: adjust resource templates

ricellis:
## Checklist

- [x] Added tests for code changes _or_ test/build only changes - build only
- [x] Updated the change log file (`CHANGES.md`) _or_ test/build only changes - build only

## Description

Adjust executor resources.

Fixes part of i350

## Approach

Align resources with scaling convention, using size 4 for Node version axes and size 5 for the LTS that also runs the agent as well as 2 test axes.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

## Monitoring and Logging

- "No change"

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -63,50 +63,6 @@ def withNpmEnv(registry, closure) {
   }
 }
 
-def nodeYaml(version) {
-    return """\
-      |    - name: node${version}
-      |      image: ${globals.ARTIFACTORY_DOCKER_REPO_VIRTUAL}/node:${version}
-      |      command: ['sh', '-c', 'sleep 99d']
-      |      imagePullPolicy: Always
-      |      resources:
-      |        requests:
-      |          memory: "2Gi"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -71,11 +71,11 @@ def nodeYaml(version) {
       |      imagePullPolicy: Always
       |      resources:
       |        requests:
-      |          memory: "2Gi"
-      |          cpu: "650m"
+      |          memory: "2490Mi"
```
emlaver, line 74: Is it worth adding a comment to ref the internal document? something like `// see 'kubePodTemplate sizing' internal pipeline document for more details`

emlaver (approved):
Looks good

eiri (approved):

Jenkinsfile
ricellis: Perhaps we should just move these templates alongside all the others, see d77f8b0

PR #744, opened 2024-06-19
Add unreliable network tests with toxiproxy

### LINKED ISSUE #360: Replace `toxy` with an alternative
## Bug Description

The toxy tests are not functioning; it looks like we'll need to replace the `toxy` package with an alternative that gets updates.

Run the `toxy` tests.

Tests to pass.

### 3. What actually happened

Tests fail with: `error happened in your connection ECONNRESET socket hang up`
which suggests the proxy is failing to start.
The `toxy` package has not had any updates for some time and it is seemingly incompatible with upstream dependency updates.

## Environment details
    > `master`
- Node.js version
    > Any

eiri:
## Checklist

## Description

This PR switches back on unreliable network tests with use of toxiproxy.

The original tests been:

| test | conditions | comment | 
| --- | --- | --- | 
| normal | No poisons to add | |
| bandwidth-limit | `poison(tpoisons.bandwidth({ bytes: 512, threshold: 1 }));` | 0.5 MB/s |
| latency | `poison(tpoisons.latency({ max: 1500, min: 250 })).withRule(trules.probability(60))` | |
| slow-read | `poison(tpoisons.slowRead({ chunk: 256, threshold: 1 })).withRule(trules.timeThreshold({ duration: 1, period: 100 }))` | Slow read for 10 % of the time e.g. 1 ms in every 100 | 
| rate-limit | `'/*/_bulk_get' poison(tpoisons.rateLimit({ limit: 20, threshold: 1000 }))` + `'/*/_bulk_docs' poison(tpoisons.rateLimit({ limit: 10, threshold: 1000 }))` | Simulate the Cloudant free plan with 20 lookups ps and 10 writes ps |

New tests are trying to match those with an exception of `rate-limit` - toxiproxy been tcp proxy can't work on HTTP level and create rules per URL or http request. Instead `bandwidth-limit` test extended to check both upstream and downstream throttling. 

Fixes #360 

## Approach

- In Jenkins for `test-network/conditions` testSuite we are switching to IAM based auth and using `http://127.0.0.1:8888` as `$COUCH_URL`
- New before block in `test-network/conditions` adds proxy on port 8080 and toxiproxy in fron of that on port 8888.
- Toxics are injected before each run, same as before.
- The run itself is `ci_e2e.js` test or backup/restore of animalsdb. The run of `largedb` excluded in Jenkins.

**Note**

Before merging Jenkins needs rollback for `RUN_TOXY_TESTS` in all blocks and set it back on 'Network Node LTS' stage.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -23,13 +23,14 @@ def getEnvForSuite(suiteName, version) {
       envVars.add("COUCHBACKUP_MOCK_SERVER_PORT=${7700 + version.toInteger()}")
       break
     case 'test-network/conditions':
+      envVars.add("CLOUDANT_IAM_TOKEN_URL=${SDKS_TEST_IAM_URL}")
       envVars.add("TEST_TIMEOUT_MULTIPLIER=50")
       envVars.add("COUCHBACKUP_MOCK_SERVER_PORT=${7800 + version.toInteger()}")
       break
-      case 'test-iam':
-        envVars.add("CLOUDANT_IAM_TOKEN_URL=${SDKS_TEST_IAM_URL}")
-        envVars.add("COUCHBACKUP_MOCK_SERVER_PORT=${7900 + version.toInteger()}")
-        break
+    case 'test-iam':
+      envVars.add("CLOUDANT_IAM_TOKEN_URL=${SDKS_TEST_IAM_URL}")
+      envVars.add("COUCHBACKUP_MOCK_SERVER_PORT=${7900 + version.toInteger()}")
+      break
     default:
       error("Unknown test suite environment ${suiteName}")
   }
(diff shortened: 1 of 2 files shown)


test-network/conditions.js
```diff
@@ -12,35 +12,123 @@
...
+      enabled: true
+    };
+    const resp = await axios.post(process.env.PROXY_URL + '/proxies', toxiProxy);
+    assert.equal(resp.status, 201, 'Should create proxy "couchdb".');
+    await waitForSocket(8888);
+  });
+
+  after('remove proxy', async function () {
+    const resp = await axios.delete(proxyURL);
+    assert.equal(resp.status, 204, 'Should remove toxic "couchdb".');
```
eiri, line 111:
```suggestion
    assert.equal(resp.status, 204, 'Should remove proxy "couchdb".');
```

test-network/conditions.js
```diff
@@ -12,35 +12,123 @@
...
+    // shutdown http proxy
+    proxy.close();
```
ricellis, line 113: Suggest adding the `done` callback so that the proxy close completes before the ater hook exits, ofc this will need a matching `function(done)` change at the function declaration. I don't think we can have the `done ` cb parameter with an `async` so we'll either need to split into 2 separate `after` hooks, or we can `promisify` the `close` function and return the promise instead.

```suggestion
    proxy.close(done);
```

ricellis (changes requested):
Looks good, a couple of small things to sort

test-network/conditions.js
eiri: Done in https://github.com/IBM/couchbackup/pull/744/commits/a400b51a91c0a009a7da7a3f927e6e5cd37ad6e8 although a bit differently to preserve `async` on teardown.

eiri:
Latest successful run of the changes: https://cloudant-sdks-jenkins.swg-devops.com/blue/organizations/jenkins/Main%2Fcouchbackup/detail/360-use-toxyproxy/32/pipeline/45

test-network/conditions.js
eiri: Fixed in https://github.com/IBM/couchbackup/pull/744/commits/2c331d722e3021648afef3805560ad1a31308404

Jenkinsfile
```diff
@@ -163,31 +165,36 @@ pipeline {
           }
         }
         stage('Node LTS') {
+          when {
+            environment name: 'RUN_TOXY_TESTS', value: 'true'
+          }
```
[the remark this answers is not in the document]
eiri: Done in https://github.com/IBM/couchbackup/pull/744/commits/149052a803958985b548ecd688f8dcbaf02f2ba1

ricellis (approved):

PR #779, opened 2024-11-15
feat: improve error response messages

emlaver:
## Checklist

## Description
Improve the error message interceptor function.  
Simplify the construction of the final error message by extracting the message from `errors` instead of from `error` and `reason` fields.

fixes i413

## Schema & API Changes
- "No change"

## Security and Privacy
- "No change"

## Testing
- Updated tests with new error message format

## Monitoring and Logging
- "No change"


includes/request.js
```diff
@@ -37,22 +37,16 @@ function errorHelper(err) {
...
-          // Override the status text with our more complete message
-          errorMsg += ` - Error: ${err.response.data.error}`;
-          if (err.response.data.reason) {
-            debug('Augmenting error message with reason property');
-            errorMsg += `, Reason: ${err.response.data.reason}`;
-          }
+        // Append the 'errors' message if available
+        if (err.response.data.errors && err.response.data.errors.length > 0) {
+          const originalError = err.response.data.errors[0];
+          errorMsg += ` - Error: ${originalError.message}`;
```
ricellis, line 49: I think we should keep any other elements that might be present and just augment the one message.
```suggestion
          errorMsg += ` - Error: ${originalError.message}`;
          originalError.message = errorMsg;
```

includes/request.js
```diff
@@ -37,22 +37,16 @@ function errorHelper(err) {
...
         }
       } else {
         errorMsg += err.response.data;
```
ricellis, line 52: suggest we move the block
```
      // Set a new message for use by the node-sdk-core
      // We use the errors array because it gets processed
      // ahead of all other service errors.
      err.response.data.errors = [{ message: errorMsg }];
```
inside this `else`  so we can accommodate the other suggestion and avoid overwriting the entire array in the other case.

ricellis (commented):
Needs a copyright update (I missed it earlier too) and a CHANGES entry.
I think we could avoid overwriting the whole array too WDYT?

emlaver:
> I think we could avoid overwriting the whole array to WDYT?

Yes, that's fine.  I originally had this solution in my branch but thought maybe we should just change the least amount of code.

ricellis (approved):

PR #810, opened 2025-01-27
test: avoid encryption key derivation warning

ricellis:
## Checklist

- [x] Added tests for code changes _or_ test/build only changes - test code
- [x] Updated the change log file (`CHANGES.md`) _or_ test/build only changes - test code

## Description

Avoid encyrption key derivation warning in test code.

Run the encryption tests.

No warnings in logs.

### 3. What actually happened

```
[2025-01-27T13:48:42.014Z] *** WARNING : deprecated key derivation used.
[2025-01-27T13:48:42.014Z] Using -iter or -pbkdf2 would be better.
```

## Approach

Change the options used on the `openssl` process to avoid the warning.
Also base64 encode the output and adjust the assertion to match.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- Modified existing tests because to avoid emitting warning and assert on base64 output.

## Monitoring and Logging

- "No change"

### DIFF
--- a/test/citestutils.js
+++ b/test/citestutils.js
@@ -1,4 +1,4 @@
-// Copyright © 2017, 2024 IBM Corp. All rights reserved.
+// Copyright © 2017, 2025 IBM Corp. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


eiri (approved):
legit

PR #957, opened 2026-06-04
fix: avoid duplicate CLI option config

ricellis:
## Checklist

- [x] Added tests for code changes _or_ test/build only changes - covered by existing tests
- [x] Updated the change log file (`CHANGES.md`) _or_ test/build only changes - refactor for tests, no user facing changes

Commander 15 update causes some tests to fail - see #956

Tests to pass.

### 3. What actually happened

CLI tests failed with
```
Error: Cannot add option '-V, --version' to command 'test' due to conflicting flag '--version'
-  already used by option '-V, --version'
```

Commander `15` enforces that options are added only once, it turns out our parsing code was adding them multiple times.
In general usage this was not an issue because the parse is only called once.
In test usage this causes a problem because parse is called repeatedly by different test combinations.

## Approach

* Create a `new Command()` for each of `backupProgram` and `restoreProgram`.
* Call the `cliDefaults()` once instead of in each parse function.
* Move the configuration of the CLI flag options out of the parse functions and onto the `const`s that define the `backup` or `restore` commands.

With these changes the options are configured only once and the tests pass, both here and with commander 15.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- No new tests because covered by existing tests

## Monitoring and Logging

- "No change"

### DIFF
--- a/includes/parser.js
+++ b/includes/parser.js
@@ -1,4 +1,4 @@
-// Copyright © 2017, 2025 IBM Corp. All rights reserved.
+// Copyright © 2017, 2026 IBM Corp. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


includes/parser.js
veronika-alraheem: Is this the copyright date you mentioned needs updating?

ricellis: Yes, now fixed in 475093c0d7c2608a9d3827590d754ab6ee0315c4

ricellis:
I also fixed the imports to the more standard ordering in 5e6bb933793d68f9895275fa24f4021716e09b2e - I had moved commander as part of the original change, but didn't see `path` lurking down the list. I thought we had some linting rules for these, but it might only be in `cloudant-node-sdk`.

veronika-alraheem (approved):

PR #960, opened 2026-06-11
ci: increase test flexibility

ricellis:
## Checklist

- [x] Added tests for code changes _or_ test/build only changes - test changes
- [x] Updated the change log file (`CHANGES.md`) _or_ test/build only changes - test changes

## Description

Use matrix testing and introduce parameter for more flexibility in activating particular test combinations.

Fixes part of i580

## Approach

See `Testing` below

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

Modified `Jenkinsfile` to support more flexible running:

- Introduced matrix-based test execution with configurable combinations of Node versions, authentication types, and test suites
- Added `shouldRunQaCombination` function to control which test combinations run via the `QA_MATRIX_INCLUDE` environment variable or build parameter
- Refactored `getEnvForSuite` to accept IAM authentication parameter and dynamically calculate port offsets
- Added `runQaCombination` to orchestrate test execution across different Node versions and test types
- Updated `runTest` to use lock resources for sequential network tests while allowing parallel execution of other tests (since we only have a single toxic proxy there)
- Improved test suite naming to include authentication type (legacy/IAM) for better test result identification
- Moved SonarQube and Mend scan stages outside the parallel execution block

This change is green with the same number of tests as main.
I've also experimented with some other test combinations

## Monitoring and Logging

- "No change"

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,10 +13,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-def getEnvForSuite(suiteName, version) {
+def getEnvForSuite(suiteName, version, iamAuth) {
 
   def envVars = []
 
+  basePort = 7700
   portOffset = 0
   switch(version) {
     case 'old-maintenance':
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -173,95 +229,108 @@ pipeline {
...
+                axis {
+                  name 'TESTS'
+                  values 'unit'
+                }
+                axis {
+                  name 'IAM_AUTH'
+                  values 'true'
+                }
+              }
+            }
```
mojito317, line 317: For me it would be cleaner to see the tests are running on each axis when `IAM_AUTH` is false OR when `IAM_AUTH` is true AND e2e AND active, instead of these excludes, but I might miss something here.

Perhaps with this simplification we could remove the shouldRunQaCombination helper as well.

Jenkinsfile
```diff
@@ -173,95 +229,108 @@ pipeline {
...
+            when {
+              beforeAgent true
+              expression {
+                shouldRunQaCombination(env.NODE_VERSION, env.IAM_AUTH, env.TESTS)
+              }
+            }
+            stages {
+              stage('Run Test') {
```
mojito317, line 325: Is not it possible to pass through here the env vars to the name of the stage? It would show a better outlook for the pipeline, especially on Blue Ocean layout.

Jenkinsfile
ricellis: The Jenkins matrix gives you all combinations of all axes.
In our case that is 3 node versions x 2 auth types x 3 test types = 18 stages

However, we have some invalid combinations:
* `network` with legacy auth (our proxy doesn't have code to handle this currently)
* `unit` with iam auth (actually this could be either way the auth makes no difference to the unit tests, but we don't want to run them twice so I had to exclude one of them)

Those 2 excludes get applied across all 3 node versions - reducing the stages from 18 to 12.

The `shouldRunQaCombination` is controlling which of those 12 are activated:
* Normal builds (matching what we do currently) 7/12 stages
    *  `unit` tests for all 3 node versions
    * `e2e` tests for all 3 node versions (legacy)
    * `e2e` tests for active node (iam)
* "big" builds 4/12 stages
    * `e2e` tests (with slow filter) for all 3 node versions (iam)
    * `network` tests for active node version (iam)

This leaves out only the `network` tests for the 2 other node versions (which could be toggled on via the parameter at any time, as indeed can any combination of _valid_ stages).

`shouldRunQaCombination` is sufficient to control the running stages without the matrix excludes, but I had two problems with it:
1. It makes 6 extra stages (that we can't use) show up in every UI view of the build so having the excludes reduces the clutter
2. It allows users to specify combinations that will fail (in this approach you can still specify them, but they can't fail because they are hard-excluded)

We could do 2 by hard-coding some excludes in `shouldRunQaCombination` but it would still clutter the UI.

>`IAM_AUTH is false` OR `IAM_AUTH is true AND e2e AND active`

This is fine for the `main` build, but not for the "big" one. The alternative would be two separate `Jenkinsfile`s but I strayed away from that on the basis of them having so much in common that'd we probably want a groovy include and so on.

Does that help explain my decision? I probably should have been more clear in the PR description about the use of `QA_MATRIX_INCLUDE` and the "big" build.

Jenkinsfile
ricellis: I already tried this when I was making the change, sadly the stage names cannot be dynamic like that, but if you hover it shows the params for the stage

image

Jenkinsfile
mojito317: Yes, thank you!
A comment around `defaultQaMatrixIncludes` or `matrixIncludes` would be nice about the "big" build.

Jenkinsfile
mojito317: It is unfortunate 😞 .

Jenkinsfile
ricellis: Improved comments in 2091b7ff1675c0b40b43b79a5cf11abb26d28a62

mojito317 (approved):

PR #967, opened 2026-06-30
fix: validate option types before log/resume validation

veronika-alraheem:
## Checklist

## Description

Avoid `(node:240) [DEP0187] DeprecationWarning: Passing invalid argument types to fs.existsSync is deprecated`
Fixes i585
## Approach

The option validations are executed within a `Promise.all()`, so `validateOptions` may not complete before `validateLogOnResume`. As a result, `fs.existsSync()` is being called with invalid arguments, which triggers the deprecation warning.  The fix splits the validations into two sequential await `Promise.all()` calls. Type checks first, then the log/resume validation, so invalid types are caught before `fs.existsSync()` is reached.
## Schema & API Changes

- "No change"
## Security and Privacy

- "No change"
## Testing

## Monitoring and Logging

- "No change"

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,4 +1,4 @@
-// Copyright © 2017, 2024 IBM Corp. All rights reserved.
+// Copyright © 2017, 2026 IBM Corp. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -54,7 +54,7 @@ function isSafePositiveInteger(x) {
  * @param {boolean} isIAM - A flag if IAM authentication been used.
  * @returns Boolean true if all checks are passing.
  */
-async function validateURL(url, isIAM) {
+function validateURL(url, isIAM) {
   if (typeof url !== 'string') {
     throw new OptionError('Invalid URL, must be type string');
   }
@@ -83,7 +83,7 @@ async function validateURL(url, isIAM) {
  * @param {object} opts - Options.
  * @returns Boolean true if all checks are passing.
  */
-async function validateOptions(opts) {
+function validateOptions(opts) {
   // if we don't have opts then we'll be using defaults
   if (!opts) {
     return true;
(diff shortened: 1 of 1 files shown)


app.js
ricellis, line 1: nit copyright update

ricellis: This comment being wrong predates your change, but it makes the contract for this function ambiguous.
The way we use this at the start of the backup or restore chain means it must return a `Promise` - the function being `async` means that happens automatically. However, your change doesn't meet the contract because it returns `Promise<void>`  for the default `undefined` value. Luckily we don't actually use the boolean value so it works anyway!
```suggestion
 * @returns {Promise<boolean>} `true` if all checks are passing, otherwise rejected.
```

app.js
```diff
@@ -208,18 +208,17 @@ async function attachmentWarnings(opts) {
...
     attachmentWarnings(opts)
-  ];
+  ]);
   if (isBackup) {
-    validations.push(
+    await Promise.all([
       shallowModeWarnings(opts),
       validateLogOnResume(opts)
-    );
+    ]);
```
ricellis: Your change is fine for solving the issue at hand, but it still leaves the checks in each group potentially happening in an undefined racy order.
I notice that all of these individual validation functions have been declared `async`, but none of them actually do any async work. I think the only reason they were made async was possibly to get them automatically wrapped in `Promise`s so they could be used in the `Promise` iterable here (which likely should have been a `then` chain anyway).

I think it might be better to remove the `async` declaration from the validation functions we call and guarantee the validation order:

```suggestion
  validateURL(url, isIAM);
  validateOptions(opts);
  attachmentWarnings(opts);
  if (isBackup) {
    shallowModeWarnings(opts);
    validateLogOnResume(opts);
  }
  return true;
```

WDYT?

ricellis (commented):
I think this should be a `fix` rather than a `refactor` commit. It is a bug fix because in some future Node version the code could be intermittently broken returning an incorrect error to the user (invalid args for existsSync instead of our own invalid option error) if the validations run out of order.

app.js
veronika-alraheem: Good point, agreed. Removed async from all the validation functions and used a fixed order:
 f98d9973957be11807726f1007bbc146daeb5cda

app.js
veronika-alraheem: Yes, I missed the return type. Fixed in: bbb7289cd06613bfeda656e00c24d4bdeb749a15

ricellis (approved):

PR #971, opened 2026-07-28
ci: add stage to auto-update peerDependencies

veronika-alraheem:
## Checklist

## Description

Added `Update peerDependencies` stage to Jenkinsfile to auto-update `cloudant-node-sdk` peer dependencies.
Fixes i596
## Approach

- Added a Jenkins stage that only runs for `cloudant-node-sdk` Dependabot update branches.
- The stage is gated by `when` conditions to ensure it only executes on the initial build and only for the expected dependency update.
- The SDK version is read from the Dependabot update and used to determine the corresponding `ibm-cloud-sdk-core` and `axios` versions.
- The peer dependencies are then updated to the versions required by the updated SDK.
## Schema & API Changes

- "No change"
## Security and Privacy

- "No change"
## Testing

## Monitoring and Logging

- "No change"

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -191,6 +191,10 @@ def runTest(version, filter, testSuite, reportName, iamAuth) {
   }
 }
 
+def sdkVersion
+def coreVersion
+def axiosVersion
+
 pipeline {
   agent {
     kubernetes {
@@ -220,6 +224,48 @@ pipeline {
           withEnv(['NPMRC_EMAIL=' + env.NPMRC_USER]) {
             withNpmEnv(registryArtifactoryDown) {
               sh 'npm ci'
+            }
+          }
+        }
+      }
+    }
+    stage('Update peerDependencies') {
+      when {
+        beforeAgent true
+        branch pattern: 'dependabot/npm_and_yarn/ibm-cloud/cloudant-\\d+\\.\\d+\\.\\d+', comparator: 'REGEXP'
+        environment name: 'BUILD_NUMBER', value: '1'
+        expression {
+          sh(returnStatus: true, script: '''
+            ! git diff HEAD~1 --name-only | grep -qvE '^package(-lock)?\\.json$' &&
+            git diff HEAD~1 -- package.json | grep '^[+-]' | grep -v '^[+-][+-][+-]' | grep -q '@ibm-cloud/cloudant'
+          ''') == 0
+        }
+      }
+      steps {
+        script {
+          sdkVersion = sh(returnStdout: true, script: "jq -r '.dependencies.\"@ibm-cloud/cloudant\"' package.json").trim()
+        }
+        withCredentials([usernamePassword(usernameVariable: 'NPMRC_USER', passwordVariable: 'NPMRC_TOKEN', credentialsId: 'artifactory')]) {
+          withEnv(['NPMRC_EMAIL=' + env.NPMRC_USER]) {
+            withNpmEnv(registryArtifactoryDown) {
+              script {
+                coreVersion = sh(returnStdout: true, script: "npm view @ibm-cloud/cloudant@${sdkVersion} dependencies.ibm-cloud-sdk-core").trim()
+                axiosVersion = sh(returnStdout: true, script: "npm view ibm-cloud-sdk-core@${coreVersion} dependencies.axios").trim()
+              }
+              sh "npm install --save-exact --save-peer axios@${axiosVersion} ibm-cloud-sdk-core@${coreVersion}"
+            }
+          }
+        }
+        sh """sed -i -E 's|"resolved": "${getRegistryArtifactoryDown()}([^"?]+)\\?dl=https%3A%2F%2F[^"]*"|"resolved": "${getRegistryPublic()}\\1"|g' package-lock.json"""
+        gitsh('github.com') {
+          script {
+            def diffStatus = sh(returnStatus: true, script: 'git diff --quiet package.json package-lock.json')
+            if (diffStatus == 1) {
+              sh 'git add package.json package-lock.json'
+              sh "git commit -m 'chore: update peerDependencies for @ibm-cloud/cloudant@${sdkVersion}'"
+              sh "git push origin HEAD:${env.BRANCH_NAME}"
+            } else {
+              echo 'No peerDependency changes to commit'
             }
           }
         }
--- a/package.json
+++ b/package.json
@@ -28,8 +28,8 @@
     "debug": "4.4.3"
   },
   "peerDependencies": {
-    "ibm-cloud-sdk-core": "5.5.0",
-    "axios": "1.18.0"
+    "axios": "1.18.0",
+    "ibm-cloud-sdk-core": "5.5.0"
   },
   "main": "app.js",
   "bin": {


Jenkinsfile
```diff
@@ -225,6 +229,45 @@ pipeline {
         }
       }
     }
+    stage('Update peerDependencies') {
+      when {
+        beforeAgent true
+        branch pattern: 'dependabot/npm_and_yarn/@ibm-cloud/cloudant-.*', comparator: 'REGEXP'
```
ricellis: I think we may as well be more specific about what we're matching here
```suggestion
        branch pattern: 'dependabot/npm_and_yarn/@ibm-cloud/cloudant-\d+\.\d+\.\d+', comparator: 'REGEXP'
```

Jenkinsfile
```diff
@@ -225,6 +229,45 @@ pipeline {
...
+        environment name: 'BUILD_NUMBER', value: '1'
+        changeset 'package.json'
```
ricellis:
```suggestion
        changeset 'package.json'
        changeset 'package-lock.json'
```

Jenkinsfile
```diff
@@ -225,6 +229,45 @@ pipeline {
...
+        expression {
+          sh(returnStatus: true, script: '''
+            ! git diff HEAD~1 --name-only | grep -qvE \'^package(-lock)?\\.json$\' &&
+            git diff HEAD~1 -- package.json | grep \'^[+-]\' | grep -v \'^[+-][+-][+-]\' | grep --quiet \'@ibm-cloud/cloudant\'
```
ricellis: A couple of minor tweaks
* single quotes don't need escaping within the triple-quoted block
* `--quiet` -> `-q` for consistency
```suggestion
            ! git diff HEAD~1 --name-only | grep -qvE '^package(-lock)?\\.json$' &&
            git diff HEAD~1 -- package.json | grep '^[+-]' | grep -v '^[+-][+-][+-]' | grep -q '@ibm-cloud/cloudant'
```

Jenkinsfile
```diff
@@ -191,6 +191,10 @@ def runTest(version, filter, testSuite, reportName, iamAuth) {
   }
 }
 
+def sdkVersion
+def coreVersion
+def axiosVersion
```
ricellis, line 196: Only `sdkVersion` needs to be scoped here I think, the others can live within their `script` block

Jenkinsfile
```diff
@@ -225,6 +229,45 @@ pipeline {
...
+      steps {
+        script {
+          sdkVersion = sh(returnStdout: true, script: "jq -r '.dependencies.\"@ibm-cloud/cloudant\"' package.json").trim()
+        }
+        withCredentials([usernamePassword(usernameVariable: 'NPMRC_USER', passwordVariable: 'NPMRC_TOKEN', credentialsId: 'artifactory')]) {
+          withEnv(['NPMRC_EMAIL=' + env.NPMRC_USER]) {
+            withNpmEnv(registryArtifactoryDown) {
+              script {
+                coreVersion = sh(returnStdout: true, script: "npm view @ibm-cloud/cloudant@${sdkVersion} dependencies.ibm-cloud-sdk-core").trim()
+                axiosVersion = sh(returnStdout: true, script: "npm view ibm-cloud-sdk-core@${coreVersion} dependencies.axios").trim()
```
ricellis, line 253:
```suggestion
                def coreVersion = sh(returnStdout: true, script: "npm view @ibm-cloud/cloudant@${sdkVersion} dependencies.ibm-cloud-sdk-core").trim()
                def axiosVersion = sh(returnStdout: true, script: "npm view ibm-cloud-sdk-core@${coreVersion} dependencies.axios").trim()
```

Jenkinsfile
```diff
@@ -225,6 +229,45 @@ pipeline {
...
+              }
+              sh "npm install --save-exact --save-peer axios@${axiosVersion} ibm-cloud-sdk-core@${coreVersion}"
+            }
```
ricellis, line 256: I think we still need an additional step here to switch host in the package-lock from artifactory to the public registry

Jenkinsfile
```diff
@@ -225,6 +229,45 @@ pipeline {
...
+          }
+        }
+        gitsh('github.com') {
+          sh """
+            if ! git diff --quiet package.json package-lock.json; then
+              git add package.json package-lock.json
+              git commit -m 'chore: update peerDependencies for @ibm-cloud/cloudant@${sdkVersion}'
+              git push origin HEAD:${env.BRANCH_NAME}
+            fi
+          """
```
ricellis: I think it would be clearer to run these as separate steps, not least because if there are no changes in package.json and package-lock.json the stage should fail, as it is we'll just skip the add/commit/push.


```suggestion
          sh '! git diff --quiet package.json package-lock.json'
          sh 'git add package.json package-lock.json'
          sh "git commit -m 'chore: update peerDependencies for @ibm-cloud/cloudant@${sdkVersion}'"
          sh "git push origin HEAD:${env.BRANCH_NAME}"
```

ricellis (changes requested):
A couple of things that need fixing, plus some other suggestions.
Overall I think this is good, but I'd like to see an actual commit; I tried it out with a fake branch, but it didn't work - I think because of the over-escaped single-quotes.

Jenkinsfile
veronika-alraheem: Should I move the `npm install` into the script block then? If `coreVersion` and `axiosVersion` aren't declared in the outer scope, won't they be unavailable to the sh step afterward?

ricellis: Ah yeah, sorry, I missed a brace there.

Just leave them global then. It would be nice if we could scope them to the stage, but I think the only way would be a stage `environment` block, but then when we set the value like `env.coreVersion` it I think gets promoted to a global anyway so we wouldn't gain anything.

Jenkinsfile
veronika-alraheem: I added the conditional because @mojito317  mentioned an edge case where Dependabot bumps `@ibm-cloud/cloudant`, but the corresponding `ibm-cloud-sdk-core` and `axios` peer dependency versions haven't changed. In that scenario there would be nothing to commit. However, I'm not sure whether that can actually happen in practice. Should the stage fail in that case as well?

ricellis: That's a good point, maybe we can return the status code then if/else on that so we can either log a warning in the build log or do the changes. That probably gives us the best of both worlds, the stage won't fail, but we'll at least have a record that there were no changes instead of a silent bypass.

Jenkinsfile
```diff
@@ -225,6 +229,50 @@ pipeline {
         }
       }
     }
+    stage('Update peerDependencies') {
+      when {
+        beforeAgent true
+        branch pattern: 'dependabot/npm_and_yarn/ibm-cloud/cloudant-\\d+\\.\\d+\\.\\d+', comparator: 'REGEXP'
+        environment name: 'BUILD_NUMBER', value: '1'
+        changeset 'package.json'
+        changeset 'package-lock.json'
```
ricellis: These didn't work in testing and they aren't really adding anything over the `expression` check

```suggestion
```

Jenkinsfile
```diff
@@ -225,6 +229,50 @@ pipeline {
...
+                axiosVersion = sh(returnStdout: true, script: "npm view ibm-cloud-sdk-core@${coreVersion} dependencies.axios").trim()
+              }
+              sh "npm install --save-exact --save-peer axios@${axiosVersion} ibm-cloud-sdk-core@${coreVersion}"
+            }
+          }
+        }
+        sh """sed -i -E 's|"resolved": "${getRegistryArtifactoryDown()}([^"?]+)\\?dl=https%3A%2F%2F[^"]*"|"resolved": "${getRegistryPublic()}\\1"|g' package-lock.json"""
+        gitsh('github.com') {
+          script {
+            def diffStatus = sh(returnStatus: true, script: 'git diff -q package.json package-lock.json')
```
ricellis: `-q` is apparently not shorthand for `--quiet` for `git`

```suggestion
            def diffStatus = sh(returnStatus: true, script: 'git diff --quiet package.json package-lock.json')
```

ricellis (changes requested):
Two more changes from testing, but otherwise this is great.

ricellis (approved):

